PR #7, opened 2017-08-16
feat(react guidelines): initial version

adrianbicanic:
This is an initial version. I guess there are some things that could be unnecessary, as well ass some things that should be added.


JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,389 @@
...
+```
+...
+├── components
+├── containers 
+...
+```
+
+- each component should have it's own folder containing it's logic in a JSX file and it's stylesheet file.
+ (If the app is (or becomes) more complex the components can be placed in additional logically separated sub-folders for better readability and organization).
+- use **camelCase** for general folder naming and **PacalCase** for React components, their stylesheet files and their containing folders
```
DarkoKukovec: Tpyo: PascalCase

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,389 @@
...
+...
+├── components
+├── containers 
+...
+```
+
+- each component should have it's own folder containing it's logic in a JSX file and it's stylesheet file.
+ (If the app is (or becomes) more complex the components can be placed in additional logically separated sub-folders for better readability and organization).
+- use **camelCase** for general folder naming and **PacalCase** for React components, their stylesheet files and their containing folders 
+- each component type's root folder should have an **index.jsx** file which exports all components, so you can later import them from a single place
```
DarkoKukovec: 1. We're not using `jsx` extensions, only `js`. When using typescript, we have to use `tsx` (compiler imitation).
2. We're doing this on some projects, but I'm not actually convinced it's a good practice, because it makes code splitting harder/impossible. Might be better to just remove this.

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,389 @@
...
+├── containers 
+    ├── Login
+        └── Login.jsx
+        └── Login.scss 
+    └── index.jsx
+...
+```
+
+### Component Declaration
+- don't use React.createClass
```
DarkoKukovec: You should probably start the sentences after a subtitle with a capital letter

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,389 @@
...
+    // ...
+    render() {
+        const {id, name} = this.props;
+        // ...
+        return (
+          <div>{name}</div>
+        );
+    }
+}
+```
```
DarkoKukovec: You should also add a TypeScript example

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,389 @@
...
+
+### Component Local State
+- if you have local state modified by the component, preffer using a single object member
+
+``` javascript
+class ComponentName extends Component {
+    // ...
+    
+    const componentState = {
```
DarkoKukovec: This example doesn't make sense:

- If you're using pure React, you would use `setState` and in this case, you would need to use the `state` property
- This is ok for MobX local state, but you should then mention this

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,389 @@
...
+            ? <ConditionallyShownComponent />
+            : null;
+        }
+    );
+}
+```
+
+### Rendering Arrays of Data
+- use array's **map** method to render items from an array
+
```
DarkoKukovec: There should always be a unique `key` property on all created elements

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,389 @@
...
+    render() {
+        return (
+            <ChildComponent
+                onClick={this.onCLickHandler}
+            />
+        );
+    }
+}
+
+// good
```
DarkoKukovec: Looks like you wanted to write something here... perhaps `@action.bound` (not sure it should be here)

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,389 @@
...
+- use camelCase for props
+- use double quotes for JSX attribute values if you are passing them a *string*
+- use self-closing tags if the element doesn't have children
+
+``` javascript
+class ComponentName extends Component {
+    // ...
+    render () {
+        return (
+          <ChildComponent
```
DarkoKukovec: You should add a separate part where you describe how to format components - if it doesn't fit into one row, put every property in a separate row and the closing `>` or `/>` in its own row.

JavaScript development/Frameworks/React.md
adrianbicanic: It's not a sentence, it's a bullet

JavaScript development/Frameworks/React.md
adrianbicanic: Added it.

JavaScript development/Frameworks/React.md
adrianbicanic: Added a TS example.

JavaScript development/Frameworks/React.md
adrianbicanic: Removed.

JavaScript development/Frameworks/React.md
DarkoKukovec: @adrianbicanic

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+}
+```
+
+## CSS Modules and Components
+- each component should have it's own stylesheet if CSS is needed for it
+
+```
+...
+├─ Login
+⎮   └─ Login.jsx
```
DarkoKukovec: This won't work :D

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+        return (
+            <Heading
+                className={styles.specialHeading}
+            />
+        );
+    }
+}
+```
+
+- if you need to pass more than one CSS class name to a component, use the [`classnames`](https://www.npmjs.com/package/classnames) library
```
DarkoKukovec: Or if you want to use conditional classes...

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+}
+```
+
+## CSS Modules and Components
+- each component should have it's own stylesheet if CSS is needed for it
+
+```
+...
+├─ Login
+⎮   └─ Login.js
```
DarkoKukovec: This still won't work :)
You have two `Login.js` files in the same folder

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+
+```
+...
+├─ components
+├─ containers 
+...
+```
+
+- use **camelCase** for general folder naming and **PascalCase** for React components, their stylesheet files and their containing folders 
+- each component type's root folder should have an **index.js** file which exports all components, so you can later import them from a single place
```
DarkoKukovec: I would remove this rule. It seems a good idea at first, but it makes code splitting very hard/impossible.

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+
+*Guides are not rules and should not be followed blindly. Use your head and think.*
+
+## File Organization and Naming
+- as it is a good practice to divide React components as container and presentational components, you should place them in separate folders (example: **containers** and **components**)
+
+```
+...
+├─ components
+├─ containers
```
andreicek: in some cases this can be pages

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+    return (
+      <div>
+        {this.componentState.message}
+      </div>
+    );
+  }
+});
+    
+// good
+import React, {Component} from 'react';
```
andreicek: I would go with `React.Component` here. Scales well with TS, and will make sense soon.

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+
+class ComponentName extends Component {
```
andreicek: I prefere `export default class ComponentName`

@DarkoKukovec feel free to chime in.

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+  return
+    <div>
+      <ChildComponent />
+    </div>;
+}
+
+// good
+render() {
+  // ...
+  return <ChildComponent />;
```
andreicek: I like this, but

```
render (
 <ChildComponent />
);
```

scales better. What do you think?

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+    id: PropTypes.string.isRequired,
+    name: PropTypes.string,
+};
+
+ComponentName.defaultProps = {
+    name: 'Default',
+};
+```
+
+- you can also declare property types as a **static member** inside a class:
```
andreicek: this is still not always possible. you need a babel plugin

JavaScript development/Frameworks/React.md
andreicek: Should we separate this in another guide?

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+
+``` javascript
+render() {
+    // ...
+    return (
+        <AlwaysShownComponent />
+        
+        {
+            isConditionMet
+            ? <ConditionallyShownComponent />
```
andreicek: missing spaces here

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+
+function ComponentName(props: IComponentProps) {
+  return (
+    // ...
+  );
+}
+```
+
+
+## Conditional Rendering
```
andreicek: this isn't a style thing, it has to be this way... does it have sense to leave it in?

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+            : null;
+        }
+    );
+}
+```
+
+## Rendering Arrays of Data
+- use array's **map** method to render items from an array
+- you should have a unique key for each element
+- **don't use array's index as a key**
```
andreicek: there is a use case, mention when it's a thing and when it's not. what about local arrays?

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+⎮   └─ Login.js
+⎮   └─ Login.js
+...
+```
+
+- importing:
+
+``` javascript
+// ...
+import * as styles from 'ComponentName.scss';
```
andreicek: incorrect path, and why the `*`?

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,489 @@
...
+    }
+}
+```
+
+- if you need to pass more than one CSS class name to a component or want to use conditional classes, use the [`classnames`](https://www.npmjs.com/package/classnames) library
+
+``` javascript
+import classNames from 'classnames';
+// ...
+import * as styles from 'ComponentName.scss';
```
andreicek: same as the previous comment.

JavaScript development/Frameworks/React.md
adrianbicanic: I commented on that above this example in the guide, and updated it to be more specific.

JavaScript development/Frameworks/React.md
adrianbicanic: Added both.

adrianbicanic:
@andreicek @DarkoKukovec I moved the TypeScript guide to it's own folder, and updated React guide.

@andreicek Please merge when everything is confirmed OK.

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+// good
+export default class ComponentName extends React.Component {
+  // ...
+}
+```
+
+If components don't have or change state, use **pure functions**:
+
+``` jsx
+function PureComponent({message}) {
```
andreicek: add an arrow function example

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+  // ...
+};
+```
+
+### Formatting
+
+Use self-closing tags, separated with an empty space, for element that don't have children.
```
andreicek: for elements that

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+
+``` jsx
+// bad
+<ChildComponent onClick={this.onClickHandler}></ChildComponent>
+
+// good
+<ChildComponent onClick={this.onClickHandler} />
+```
+
+If props don't fit on one line separate them with line breaks and also, put opening and closing tags in separate lines.
```
andreicek: If there is more than one prop

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+
+  render () {
+    return (
+      <ChildComponent onClick={this.onClickHandler} />
+    );
+  }
+}
+```
+
+Omit values for boolean props if default value is **true**.
```
andreicek: Move this before the functions

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+  onClick={this.onClickHandler}
+  isPropActive
+/>
+```
+
+### Receiving Props
+*Checkout [`prop-types`](https://github.com/facebook/prop-types) lib on GitHub for general info.*
+
+In class components:
+- always use prop validation
```
andreicek: mention typescript here

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+  render() {
+    const {id, name, onClickHandler} = this.props;
+    // ...
+  }
+}
+```
+
+### Importing React
+``` jsx
+import * as React from 'react';
```
andreicek: This will work only in TS. Babel is `import React from 'react';`

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+
+ComponentName.propTypes = {
+  id: PropTypes.number.isRequired,
+  name: PropTypes.string
+}
+```
+
+### Using props
+
+If you are using more than one property in a method, preffer accessing them with **object destructuring**.
```
andreicek: Mention also extracting complex variables in scope instead of defining them in JSX

```
render() {
    const trasnlatedLabel = label || this.t('translation');
    
    return (
           <Component label={trasnlatedLabel} />
    );
}
```

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+// exporting when static members are not supported by the transpiler:
+class ComponentName extends React.Component {
+  // ...
+}
+
+ComponentName.propTypes = {
+  // ..
+};
+
+export default ComponentName;
```
andreicek: Separate in two code blocks.

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+}
+
+PureComponent.propTypes = {
+  // ...
+};
+
+export default PureComponent;
+
+// stateless components with no prop validation
+export default () => (/* ... */);
```
andreicek: Same

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+```
+
+### Rendering Arrays of Data
+
+For iterating arrays:
+- use **map** method to render items
+- you should have a unique key for each element
+- **don't use array's index as a key**
```
andreicek: Please mention when this is ok. What is book doesn't have an id? What will you do then?

@DarkoKukovec chime in.

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+    );
+  }
+}
+```
+
+## React + MobX
+*For general information about MobX, checkout the official documentation:* [`mobx.js.org`](https://mobx.js.org/).
+
+## Decorators
+Use decorators wherever you can because it makes code more readable and declarative.
```
andreicek: Same as static you may not have support for this.

andreicek:
Please update the documentation to forbid default exporting.

JavaScript development/Frameworks/React.md
DarkoKukovec: This happens if you use indexes: https://twitter.com/ryanflorence/status/920715795088384000
They're only acceptable if you don't have any choice and the list is read-only

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+- **don't use array's index as a key**
+
+``` jsx
+render() {
+  // ...
+  return (
+    <ul>
+      {
+        booksArray.map((book) => {
+          <ListItem key={book.id} title={book.title} />
```
DarkoKukovec: There should also be a different example - an example where the items have nested elements, e.g. a link inside of a `<li>`

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+
+  render() {
+    return (
+      <ChildComponent onClick={this.onClickHandler.bind(this)} />
+    );
+  }
+}
+    
+// good
+class ComponentName extends React.Component {
```
DarkoKukovec: Maybe we should think about using something like https://github.com/andreypopp/autobind-decorator
@andreicek

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+import classNames from 'classnames';
+// ...
+import styles from './ComponentName.scss';
+
+class ComponentName extends React.Component {
+  // ...
+  render() {
+    return (
+      <Heading
+        className={classNames(styles.specialHeading, styles.largeHeading)}
```
DarkoKukovec: Also add an example with conditional classes

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+@observer
+class ComponentName extends React.Component {
+  // ...
+}
+```
+
+### Local Component State
+If MobX is used for state management:
+- preffer using MobX *observable* instead of *this.state* for component's local state; don't use both
+- wrap local state in a single object just as it would be in the built in state
```
DarkoKukovec: Make it clear that some name other than `state` should be used. It might also make sense to define a name here, so we're using the same name everywhere. @andreicek

JavaScript development/Typescript/TypeScript.md
```diff
@@ -0,0 +1,117 @@
...
+  isConditionMet: boolean;
+  onClick?: (event: any) => void;
+}, {}> {
+  // ...
+}
+```
+
+In stateless components, use regular object destructuring with TypeScript's type annotations.
+``` typescript
+export default ({className?: string, title: string}): JSX.Element => {
```
DarkoKukovec: This is wrong. In this example `string` is not a type, but a variable (default value). Correct example:
`export default ({className, title}: {className?: string, title: string}): JSX.Element => {`

JavaScript development/Typescript/TypeScript.md
```diff
@@ -0,0 +1,117 @@
...
+    // ...
+}
+```
+
+Use interfaces for props when you have a great amount of props/state to validate or when you have a set of common props used in more than a few components (in which case the interfaces should be imported from a separate file).
```
DarkoKukovec: We should remove the "great amount of props/state to validate" case, and only use interfaces if they're shared across multiple components.

JavaScript development/Typescript/TypeScript.md
```diff
@@ -0,0 +1,117 @@
...
+class ComponentName extends Component<{}, {}> {
+  // ...
+}
+```
+
+### Class Members and Methods
+React's lifecycle hooks and render method have to be declared public. Local data and methods that are not passed to child components are generally declared private.
+
+``` typescript
+class ComponentName extends Component<IComponentNameProps, IComponentNameState> {
```
DarkoKukovec: Remove interfaces from here, they're not essential for the example, so `{}` should be used.

JavaScript development/Frameworks/React.md
andreicek: I really like this idea. How would do this with `@action.bound`? Would that still be the preferred way of binding actions or would we do this:

```javascript
@autobind
@action
doStuff() { /* code */ }
```

Here is a working example of this: https://codesandbox.io/s/v3jlrpy3w7

I actually think this is better since the methods usually start off not being actions.

andreicek: There is a known problem with `autobind-decorator` and TS. In the issue someone linked developit/decko module that has bind.

Here is an example of this running: https://codesandbox.io/s/ox0p2pym7z

DarkoKukovec: Looks good! Would avoid the `memoize` function trough, seems like it could cause unexpected bugs based on the way it caches the results.

andreicek: Agreed! I would then mention this lib and prefere it instead of `this.fn = this.fn.bind(this);`

@adrianbicanic what do you think?

adrianbicanic: I can comment only regarding how the code looks. Def looks cleaner and more consistent than stuffing 6 methods bindings in the constructor and then binding some more action methods with decorators.

andreicek: OK - let it be so.

- use `developit/decko` lib
- `@bind` over constructor binding
- `@bind` over `@action.bound`
- don't use `memoize`

andreicek:
@DarkoKukovec please have a look and merge if ok.

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,598 @@
...
+      <ChildComponent onClick={this.onClickHandler.bind(this)} />
+    );
+  }
+}
+
+// bad
+export class ComponentName extends React.Component {
+  constructor(args) {
+    super(args);
+    this.onClickHandler = this.onClickHandler.bind(this);
```
DarkoKukovec: Wouldn't say this is necessary bad. This is the recommended way if you don't have decorators available (they're not a standard feature, we're adding them because of mobx)

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,598 @@
...
+
+export {FunctionalComponent};
+```
+
+### Rendering Arrays of Data
+
+For iterating arrays:
+- use **map** method to render items
+- you should have a unique key for each element
+- **don't use array's index as a key**, unless you have no choice (no unique property in each element of the array)
```
DarkoKukovec: I would also add that the elements need to keep their positions (constant indexes). If not, this causes some very unpredictable bugs.

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,598 @@
...
+
+  render () {
+    return (
+      <ChildComponent onClick={this.onClickHandler} />
+    );
+  }
+}
+```
+
+Avoid writing any logic inline. Extract it to the outer scope.
```
DarkoKukovec: "outer scope" sounds confusing. Maybe better to change it to something like "outside of the JSX block"

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,598 @@
...
+/>
+
+// good
+<ChildComponent
+  onClick={this.onClickHandler}
+  isPropActive
+/>
+```
+
+Declare functions before passing them as props instead of passing inline anonymous (arrow) functions. They will be needlessy re-rendered each time, thus effecting performance.
```
DarkoKukovec: "re-rendered" is the wrong word here. Functions are not rendered, a better word might be "recreated"

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,598 @@
...
+⎮       └─ Login.js
+⎮       └─ Login.scss 
+...
+```
+
+### Component Declaration
+
+For declaring class components:
+- use ES2015 **class syntax**
+- don't use *React.createClass*
```
DarkoKukovec: Maybe add a note:
If you absolutely have to, use the [`create-react-class`](https://www.npmjs.com/package/create-react-class) package instead of `React.createClass` (removed in React 16).

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,598 @@
...
+import classNames from 'classnames';
+// ...
+import styles from './ComponentName.scss';
+
+export class ComponentName extends React.Component {
+  // ...
+  render() {
+    return (
+      <Heading
+        className={classNames(styles.largeHeading, {
```
DarkoKukovec: Should we also follow our rule here of extracting the logic from JSX?

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,598 @@
...
+          [styles.importantHeading]: isImportant,
+        })}
+      />
+    );
+  }
+}
+```
+
```
DarkoKukovec: Maybe add one more item here defining the max number of lines in a single file? @andreicek

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,598 @@
...
+        <div>{this.componentState.counter}</div>
+        <button onClick={this.increaseCounter}>Increase by One</button>
+      </div>
+    );
+  }
+}
+```
+
+### Async Await
+In actions implemented as *async* functions, after each *await*, any code that modifies the state should be wrapped in a *runInAction* utility. This is because *async await* is just syntax sugar, and in reallity, a new function is created after each *await*.
```
DarkoKukovec: > This is because *async await* is just syntax sugar, and in reallity, a new function is created after each *await*.

This is only true for the transpiled code, but is also not relevant. The reason is that actions need to be synchronous.

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,598 @@
...
+  import {observer} from 'mobx-react';
+// ...
+export class ComponentName extends React.Component {
+  // ...
+  @observable componentState = {
+    userData: null,
+  };
+
+  @bind
+  @action
```
DarkoKukovec: The question is should we add an `@action` decorator if it's not needed (like in this case), but the function contains `runInAction`? @andreicek

JavaScript development/Typescript/TypeScript.md
```diff
@@ -0,0 +1,119 @@
...
+``` typescript
+export class ComponentName extends React.Component<{/*prop types*/}, {/*state types*/}> {
+    // ...
+}
+```
+
+Declare functional components as arrow functions
+
+``` typescript
+export const FunctionalComponent = (/*props*/): JSX.Element => {
```
DarkoKukovec: Is the return type valid in all cases? `null` is a valid use case (when we decide not to render a component), but I'm not sure it's assignable to `JSX.Element`.

JavaScript development/Typescript/TypeScript.md
```diff
@@ -0,0 +1,119 @@
...
+}
+``` 
+
+``` typescript
+import {CommonProps} from 'interfaces/CommonProps';
+
+export class ComponentName extends React.Component<CommonProps, {}> {
+  // ...
+}
+```
```
DarkoKukovec: Maybe also add an example when the component needs `CommonProps`, but also has something extra:
```typescript
export class ComponentName extends React.Component<CommonProps & {active?: boolean}, {}> {
  // ...
}
```

JavaScript development/Typescript/TypeScript.md
```diff
@@ -0,0 +1,119 @@
...
+
+  public render() {
+    return {
+      // ...
+    };
+  }
+}
+```
+
+Example with MobX:
```
DarkoKukovec: Is this example relevant? I don't see anything here that wasn't shown in previous examples.

DarkoKukovec:
Should we also add a Resources section with good tutorials (TS, MobX, etc)?

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,598 @@
...
+
+``` jsx
+// bad
+<ChildComponent onClick={this.onClickHandler}></ChildComponent>
+
+// good
+<ChildComponent onClick={this.onClickHandler} />
+```
+
+If there is more than one prop, separate them with line breaks and also, put opening and closing tags in separate lines.
```
dingo-d: Grammar fix: If there is more than one prop, separate them with line breaks, and also put opening and closing tags on separate lines.

JavaScript development/Frameworks/React.md
dingo-d: It's _affecting_ performance, not effecting ;)

DarkoKukovec: That was my first thought also, but no:
affecting - touching the emotions; moving.
effecting - cause (something) to happen; bring about.

DarkoKukovec: Actually, I'm not sure... would it be "affecting performance" or "effecting bad performance"?

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,598 @@
...
+    );
+  }
+}
+```
+
+## React + MobX
+*For general information about MobX, checkout the official documentation:* [`mobx.js.org`](https://mobx.js.org/).
+
+### Decorators
+Use decorators wherever you can if your transpiler support's it, because it makes code more readable and declarative.
```
dingo-d: Grammar fix:  ... if your transpiler supports it

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,598 @@
...
+    return (
+      <div>{this.fullName}</div>
+    );
+  }
+}
+```
+
+### Actions
+Wrap methods and functions that modify state with *@action*. For binding action methods, use the aforementioned *@bind* decorator from the [`developit/decko`](https://github.com/developit/decko) lib.
+``` jsx
```
dingo-d: Add new line

JavaScript development/Frameworks/React.md
dingo-d: I think this should be 'affect' because it's a verb, it affects the performance. But 'can have an adverse effect on the performance' could also be used.
Or it could be reformulated to be that 'the effect of passing inline anonymous functions is decreased performance' (sounds bad, and shouldn't be used, but just as an example for when 'effect' can be used :D).

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,536 @@
...
+In stateless components, use PropTypes for type checking but use native JS for setting default values.
+
+``` jsx
+function ComponentName({id, name = 'John Doe'}) {
+  // ...
+};
+
+ComponentName.propTypes = {
+  id: PropTypes.number.isRequired,
+  name: PropTypes.string
```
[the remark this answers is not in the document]
andreicek: fixed.

JavaScript development/Frameworks/React.md
andreicek: "instanced" is even better :)

JavaScript development/Frameworks/React.md
andreicek: If the project is not using mobx it can still use `bind` decorator so in this case, I would consider this an anti-pattern.

JavaScript development/Frameworks/React.md
andreicek: eh not sure about this since it's not business logic. your call

JavaScript development/Frameworks/React.md
andreicek: i'm not sure why `action` was used here in the first place.

JavaScript development/Typescript/TypeScript.md
andreicek: removed func. components

JavaScript development/Typescript/TypeScript.md
andreicek: yup - removed

andreicek:
you can check it again @DarkoKukovec

JavaScript development/Frameworks/React.md
```diff
@@ -0,0 +1,548 @@
...
+⎮       └─ Login.js
+⎮       └─ Login.scss 
+...
+```
+
+### Component Declaration
+
+For declaring class components:
+- use ES2015 **class syntax**
+- if you **need** `React.createClass` use [`create-react-class`](https://www.npmjs.com/package/create-react-class) (`createClass` was removed in Reactr 16)
```
DarkoKukovec: `Reactr` typo

JavaScript development/Frameworks/React.md
DarkoKukovec: But what if we don't have decorators available?

JavaScript development/Typescript/React.md
```diff
@@ -0,0 +1,78 @@
...
+  isConditionMet: boolean;
+  onClick?: (event: any) => void;
+}, {}> {
+  // ...
+}
+```
+
+In stateless components, use regular object destructuring with TypeScript's type annotations.
+``` typescript
+export const FunctionalComponent = ({classname, title}: {className?: string, title: string}): JSX.Element => {
```
DarkoKukovec: But should we use functional components at all?

PR #12, opened 2018-04-12
Feature/react short circuit operator

mamihelj09:


JavaScript/React.md
```diff
@@ -313,6 +313,16 @@ render() {
...
 ```
 
+### Conditional Rendering
+When you want to render either something or nothing, you can use the && operator.
+```
+// bad
+{condition ? (<span />) : null}
+
+// good
+{condition && (<span />)}
```
DarkoKukovec: Also need to add a comment that you need to be sure the condition returns a boolean or weird things could happen.

JavaScript/React.md
```diff
@@ -313,6 +313,16 @@ render() {
...
+```
+
```
DarkoKukovec: Also add a multiline example.

andreicek: why would you wrap `span` in `()`?

PR #26, opened 2019-01-22
Git chapter

mjanjic01:


Git.md
```diff
@@ -0,0 +1,80 @@
...
+## Flow
+
+We use the [Git Flow](https://nvie.com/files/Git-branching-model.pdf) workflow model.
+
+### Master
+The `master` branch is one of the two branches with an infinite lifetime. The source code of `HEAD` always reflects a production-ready state.
+
+### Development
+
+`development` is the second branch with an infinite lifetime. The `development` branch source code of `HEAD` always reflects a state with the latest delivered development changes for the next releases.
```
DarkoKukovec: This should be `develop`, right? That's the name used in git flow and the name we use on all (or almost all) projects.

Git.md
```diff
@@ -0,0 +1,80 @@
...
+
+When the source code in the `development` branch reaches a stable point and is ready to be released, all of the changes should be merged back into `master` and tagged with a release number.
+
+Usually, an automatic build and deploy to production server hook is triggered everytime a commit is pushed to `master`.
```
DarkoKukovec: This depends on the project, but on some projects, the deploy might not be automated and it needs to be triggered manually (from semaphore). This is most commonly done to have better control of when something is deployed so the backend and frontend deploy can be syncronised as necessary.

Git.md
mjanjic01: The name used in git flow is different in each article. You will find `develop`, `development` or just `dev` for the development branch name. Here I went with `development` for the sake of consistency, keeping all branch names nouns.

Git.md
mjanjic01: `Usually` it works like that. Deployment specifics can be added in the Deployment chapter

Git.md
DarkoKukovec: Ok, but we should go with the name we use most often, and that is `develop`

safo6m: Conclusion after offline discussion:
as we are still (kind-of) using git flow approach and most of our projects are using `develop` branch, we decided to stick with `develop` naming.

Feel free to open an issue if you are not convinced that we have made the right decision.

dingo-d:
You can look at https://handbook.infinum.co/books/rails/Git%20Process and https://handbook.infinum.co/books/wordpress/Versioning%20in%20WordPress/Versioning for reference (if needed ofc), plus, you can leave comments for any of those in their respective repos if you find something wrong/typo etc. 🙂

Git.md
```diff
@@ -0,0 +1,82 @@
...
+`develop` is the second branch with an infinite lifetime. The `develop` branch source code of `HEAD` always reflects a state with the latest delivered development changes for the next releases.
+
+When the source code in `develop` branch reaches a stable point and is ready to be released, all of the changes should be merged back into `master` and tagged with a release number.
+
+### Feature
+
+Next to the main `master` and `develop` branches the Flow model uses a variety of supporting branches to aid parallel development between team members, ease tracking of features, prepare for production releases, and to assist in quickly fixing live production problems.
+Unlike the two main branches, these branches have a limited lifetime.
+
+Feature branches may branch of from `develop` and must be merged back into `develop`. Feature branch naming convention is as follows: `feature/*`, for example `feature/login-page`
```
DarkoKukovec: Maybe add a suggestion to put the task number into the branch name if possible?

Git.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+### Commit messages
+
+Writing good commit messages is important. A clear commit log will be easier to navigate through and make it easier for other people to follow along. You can read on how to write proper [here](https://chris.beams.io/posts/git-commit/).
+
+
+## Pull requests
+
+Avoid creating pull requests that contain a lot of changes. If you are implementing a feature that requires a lot of file changes, try breaking it up into multiple smaller features which results in multiple pull requests.
```
DarkoKukovec: Maybe link the test "a lot of changes" to https://www.urbandictionary.com/define.php?term=KPR ? 😄

PR #34, opened 2019-03-10
Lektura

MirelaZ:
Stranica lektorirana

### DIFF
--- a/JavaScript/Angular.md
+++ b/JavaScript/Angular.md
@@ -1,14 +1,14 @@
-# Angular Guidelines and Best Practices
+# Angular guidelines and best practices
 
 <center><img style="width: 100%; max-width: 1000px" src="/img/angular.svg"></center>
 
 *Guides are not rules and should not be followed blindly. Use your head and think.*
 
-## Core Libraries, Configuration and Tools
+## Core libraries, configuration and tools
 
 ### [Angular CLI](https://github.com/angular/angular-cli)
 
-The best way to create a new project is by using Angular CLI. It will hide a lot of configuration behind the scenes (you will not have a webpack config exposed for editing), it does some optimizations when compiling your code and it offers some handy scaffolding tools.
+The best way to create a new project is by using Angular CLI. It will hide a lot of configuration behind the scenes (you will not have a webpack config exposed for editing), it does some optimizations when compiling your code, and it offers some handy scaffolding tools.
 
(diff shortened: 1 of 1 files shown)


JavaScript/Angular.md
```diff
@@ -969,134 +969,134 @@ class MyComponent implements OnInit, OnChanges {
...
 
 DI is utilized heavily in Angular, primarily for providing instances of either primitive values or class instances.
 
-DI is a very complex topic and warrants a separate discussion which would be out of scope for this Handbook. Make sure to check out the [official DI guide](https://angular.io/guide/dependency-injection) which explains many details. You can also have a look at [this repo](https://github.com/fvoska/angular-di-demo) for some examples.
+DI is a very complex topic and requires a separate discussion which would be out of the scope of this handbook. Make sure to check out the [official DI guide](https://angular.io/guide/dependency-injection) which explains many details. You can also have a look at [this repo](https://github.com/fvoska/angular-di-demo) for some examples.
 
 As of Angular 6, make sure to use `{ providedIn: 'root' }` [for singletons](https://angular.io/guide/singleton-services#providing-a-singleton-service) whenever possible.
 
-DI is very useful for testing purposes, as shown in next chapter.
+DI is very useful for testing purposes, as shown in the next chapter.
```
fvoska: I actually just noticed that I reordered some chapters after writing this line, so the chapter about testing is no longer immediately after DI chapter.

```suggestion
DI is very useful for testing purposes, as shown in a later chapter.
```

JavaScript/Angular.md
```diff
@@ -847,19 +847,19 @@ const userName$: Observable<string> = http.get('/account/me').pipe(
 userName$.subscribe(console.log); // TheDude
 ```
 
-The reason why is because there might be multiple subscribers who want the same thing and they would all have to do the same transformation in their respective success callbacks.
+The reason why is because there might be multiple subscribers who want the same thing, and they would all have to do the same transformation in their respective success callbacks.
 
-It is of course possible that subscribers want different things, and for those cases you might want to expose some observable which has less piped operations.
+It is, of course, possible that the subscribers want different things, and for those cases you might want to expose some observable whose operations are less piped.
```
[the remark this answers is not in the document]
MirelaZ: You're right, the meaning did change after proofreading, sorry, I thought you had something else in mind... Fewer piped operations is correct!

JavaScript/Angular.md
MirelaZ: Using "in a later chapter" is OK.

PR #37, opened 2019-03-11
Lektura

MirelaZ:
Stranica lektorirana

### DIFF
--- a/JavaScript/React.md
+++ b/JavaScript/React.md
@@ -1,9 +1,9 @@
-# React Guidelines and Best Practices
+# React guidelines and best practices
 
 *Guides are not rules and should not be followed blindly. Use your head and think.*
 
-### File Organization and Naming
-As it is a good practice to divide React components as container and presentational components, you should place them in separate folders (this example uses **containers** and **components** for folder naming, but there are other naming conventions, depending on the use case and preference of developers).
+### File organization and naming
+As it is good practice to divide React components into container and presentational components, you should place them in separate folders (this example uses **containers** and **components** for folder naming, but there are other naming conventions, depending on the use case and preference of developers).
 
 ```
 ...
@@ -12,7 +12,7 @@ As it is a good practice to divide React components as container and presentatio
 ...
 ```
 
-Use **camelCase** for general folder naming and **PascalCase** for React components, their stylesheet files and their containing folders.
+Use **camelCase** for general folder naming, and **PascalCase** for React components, their style sheet files, and their containing folders.
 
 ```
 ...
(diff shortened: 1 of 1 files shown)


JavaScript/React.md
```diff
@@ -210,7 +210,7 @@ class ComponentName extends React.Component {
 }
 ```
 
-Avoid writing any logic inline. Extract it to the outside of the JSX block.
+Avoid writing any logic in line. Extract it to the outside of the JSX block.
```
kristian240: Think it should stay "inline"

JavaScript/React.md
```diff
@@ -355,8 +355,8 @@ Make sure condition is boolean, if not cast it to boolean.
 ```
 
 ### Binding
-- don't bind functions inline or in a class constructor if you have decorators
-- use [`developit/decko`](https://github.com/developit/decko) lib's *@bind* decorator for binding
+- don't bind functions in line or in a class constructor if you have decorators
```
kristian240: inline

JavaScript/React.md
```diff
@@ -481,10 +481,10 @@ export class ComponentName extends React.Component {
 }
 ```
 
-### Local Component State
+### Local component state
 If MobX is used for state management:
-- prefer using MobX *observable* instead of *this.state* for component's local state; don't use both
-- wrap local state in a single observable object named **componentState**
+- prefer using a MobX *observable* instead of *this.state* for a component's local state; don't use both
```
kristian240: Not sure for the "a MobX" change

JavaScript/React.md
MirelaZ: I don't think inline can be used as an adverb. If it should stay inline, and inline is an adjective, would it make sense to change places and write; Avoid writing any inline logic?

JavaScript/React.md
MirelaZ: There should be an article before MobX if the sentence is structured like this - _a_ if _observable_ is not a single item, _the_ if it is a specific one item. If you don't want to treat _observable_ as a noun, I'd recommend writing, for example, prefer using MobX's _observable_ or prefer using _observable_ from MobX or something along those lines

JavaScript/React.md
kristijanbambir: I would change it to that 👍

JavaScript/React.md
kristijanbambir: As before, I propose to change this to "don't bind inline functions".

andreicek:
Closing this as the file does not exist anymore.

PR #46, opened 2019-09-26
Feature/hooks

DarkoKukovec:


JavaScript/React.md
```diff
@@ -509,10 +559,33 @@ export class ComponentName extends React.Component {
 }
 ```
 
+#### Functional components
+
+In functional components it's fine to use either `useState` or `useLocalStore`, but don't mix them together - choose one for the project and then stick with it.
+`useLocalStore` might be simpler in the short run, but it might not work with new React features like concurrent rendering.
```
DarkoKukovec: Not sure if we should leave this to a developers choice or should we choose.

dingo-d: Just add: `we recommend using...`

JavaScript/React.md
```diff
@@ -1,14 +1,17 @@
 # React Guidelines and Best Practices
 
-*Guides are not rules and should not be followed blindly. Use your head and think.*
+_Guides are not rules and should not be followed blindly. Use your head and think._
+
+If you're using Prettier, there are some styling rules here that will be overriden by the formatter. In that case, ignore the styleguides mentioned here.
```
DarkoKukovec: Just a disclaimer until we update the handbook with prettier style tweaks.

JavaScript/React.md
```diff
@@ -1,14 +1,17 @@
 # React Guidelines and Best Practices
```
dingo-d: [There are no best practices](https://www.satisfice.com/blog/archives/5164) 😬

DarkoKukovec: Didn't read the whole article yet, but I'm not sure they're talking about practices or rules - because one is a recommendation and the other is a must. We make this clear with the first sentence after the title :D

dingo-d: I was referring to the fact that there are no _best practices_ because there are no universal best practices that all should abide by.
Just the recommendations in a given situation that yield the most benefit 🙂 

I'd rename the chapter to something like 

`React Guidelines and Recommended Infinum Practices`

JavaScript/React.md
isBatak: Maybe we can add this link somewhere https://mobx-react.js.org/

JavaScript/React.md
```diff
@@ -530,20 +603,37 @@ export class ComponentName extends React.Component {
...
+    </div>
+  );
+});
 ```
 
 ### Actions
-Wrap methods and functions that modify state with *@action*. For binding action methods, use the aforementioned *@bind* decorator from the [`developit/decko`](https://github.com/developit/decko) library.
 
-``` jsx
+Wrap methods and functions that modify state with _@action_. For binding action methods, use the aforementioned _@bind_ decorator from the [`developit/decko`](https://github.com/developit/decko) library (for class components).
```
isBatak: I would go with `@action.bound` in that case...

JavaScript/React.md
```diff
@@ -574,14 +664,32 @@ export class ComponentName extends React.Component {
...
+});
 ```
 
 ### Async Await
-In actions implemented as *async* functions, after each *await*, any code that modifies the state should be wrapped in a *runInAction* utility.
 
-``` jsx
-  import { action, observable, runInAction } from 'mobx';
-  import { observer } from 'mobx-react';
+In actions implemented as _async_ functions, after each _await_, any code that modifies the state should be wrapped in a _runInAction_ utility.
```
isBatak: or use `flow`

JavaScript/React.md
```diff
@@ -324,42 +354,50 @@ render() {
...
-    // ...some stuff
-  </div>
+const condition = 1;
+{
+  Boolean(condition) && <div>// ...some stuff</div>;
+}
+
+// good
+{
+  condition && <div>// ...some stuff</div>;
```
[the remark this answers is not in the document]
DarkoKukovec: ok, it should be clearer that the condition in this case is a boolean

JavaScript/React.md
```diff
@@ -23,21 +26,23 @@ Use **camelCase** for general folder naming and **PascalCase** for React compone
 ⎮   └─ LoginHeader
```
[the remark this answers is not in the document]
DarkoKukovec: I'm not sure - on the one side, it makes sense, but on the other side it is more rules that the developer needs to keep track of

JavaScript/React.md
```diff
@@ -51,25 +56,28 @@ import logo from '../assets/logo.svg';
 ### Component Declaration
 
 For declaring class components:
-- use ES2015 **class syntax**
-- if you **need** `React.createClass` use [`create-react-class`](https://www.npmjs.com/package/create-react-class) (`createClass` was removed in React 16)
 
-``` jsx
+- use functional components for presentational components
+- use ES2015 class syntax for containers
```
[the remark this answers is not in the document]
DarkoKukovec: Good question, not sure :D

JavaScript/React.md
```diff
@@ -324,42 +354,50 @@ render() {
...
+  condition && <div>// ...some stuff</div>;
 }
 ```
 
 ### Binding
+
 - don't bind functions inline or in a class constructor if you have decorators
-- use [`developit/decko`](https://github.com/developit/decko) lib's *@bind* decorator for binding
+- use [`developit/decko`](https://github.com/developit/decko) lib's _@bind_ decorator for binding
 - don't use arrow functions as class properties
```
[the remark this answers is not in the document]
DarkoKukovec: It's ugly :'(

JavaScript/React.md
DarkoKukovec: the reason we initially said no `@action.bound` is because we noticed people being lazy and using it even if the action was not needed.

JavaScript/React.md
```diff
@@ -596,9 +704,34 @@ export class ComponentName extends React.Component {
...
+  render() {
+    /* ... */
+  }
 }
 ```
 
+For async operations in functional components, the simplest way is to use the [`useAsync`](https://github.com/streamich/react-use/blob/master/docs/useAsync.md) hook from the [`react-use`](https://github.com/streamich/react-use) package.
+
+```jsx
+import { useAsync } from 'react-use';
```
isBatak, line 716: I would avoid using `react-use` library, It's not so good...

isBatak: maybe `react-swr`

PR #49, opened 2019-11-20
Update links

cilim:
All links that redirect to another chapter inside the Handbook are now parameterized. Both styles will work, but this looks cleaner :)


cilim:
@kristijanbambir please merge this in if you're ok with the change :)

PR #51, opened 2020-01-09
Add more angular topics

fvoska:
- best practices using ngOnChanges
- different initial assignment options

### DIFF
--- a/JavaScript/Angular.md
+++ b/JavaScript/Angular.md
@@ -1006,7 +1006,7 @@ class MyComponent {
 }
 ```
 
-Property assignments will be executed immediately after constructor execution has finished, so in this example the final value oif `foo` will be `bar`.
+Property assignments will be executed immediately after constructor execution has finished, so in this example the final value of `foo` will be `bar`.
 
 We recommend setting initial values alongside property declaration since it is the shortest option.
 
--- a/JavaScript/Angular.md
+++ b/JavaScript/Angular.md
@@ -1045,7 +1045,7 @@ class PersonDetailsComponent implements OnInit {
 
 Using `ngOnInit` works and we get no exception (with the assumption that a valid `Date` object is passed down to our component via `birthDate` input).
 
-We still have one problem with this solution. If input binding changes, `isLegalAge` will not be re-assigned. To solve this we simply switch to useing `ngOnChanges`:
+We still have one problem with this solution. If input binding changes, `isLegalAge` will not be re-assigned. To solve this we simply switch to using `ngOnChanges`:
 
 ```ts
 // good
--- a/JavaScript/Angular.md
+++ b/JavaScript/Angular.md
@@ -795,6 +795,384 @@ export class AuthorModel extends User {
 
 ----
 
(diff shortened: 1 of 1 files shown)


JavaScript/Angular.md
```diff
@@ -795,6 +795,384 @@ export class AuthorModel extends User {
...
+@Component({
+  template: `
+  <ng-container *ngIf="user">
+    {{ user.name }}
+  </ng-container>
+  `,
+  ...
+})
+class MyComponent {
+  public user: UserModel = this.route.params.pipe(switchMap((params: Params) => {
```
DarkoKukovec: `user$`?

JavaScript/Angular.md
```diff
@@ -795,6 +795,384 @@ export class AuthorModel extends User {
...
+  @Input() public input2: string;
+  private upperCaseInput1: string;
+
+  public ngOnChanges() {
+    this.upperCaseInput1 = this.input1.toUpperCase();
+  }
+}
+```
+
+Problem with not checking which specific change was triggered is that whatever computation is done might be done unnecessarily. You should add checks so that the action is executed only when the related inputs change and ignore changes of other inputs. In above example `upperCaseInput1`'s value depends only on `input1`'s value, but the assignment will be executed even if only `input2` changes, which is unnecessary.
```
DarkoKukovec: Should this be "The problem"?

JavaScript/Angular.md
```diff
@@ -795,6 +795,384 @@ export class AuthorModel extends User {
...
+
+  ngOnInit() {
+    this.isLegalAge = new Date().getYear() - this.birthDate.getYear() > 18;
+  }
+}
+```
+
+Using `ngOnInit` works and we get no exception (with the assumption that a valid `Date` object is passed down to our component via `birthDate` input).
+
+We still have one problem with this solution. If input binding changes, `isLegalAge` will not be re-assigned. To solve this we simply switch to useing `ngOnChanges`:
```
kristijanbambir:
```suggestion
We still have one problem with this solution. If input binding changes, `isLegalAge` will not be re-assigned. To solve this we simply switch to using `ngOnChanges`:
```

JavaScript/Angular.md
kristijanbambir: Actually, no, it should stay `user`, but initial assignment should be removed.

JavaScript/Angular.md
```diff
@@ -795,6 +795,384 @@ export class AuthorModel extends User {
...
+  }));
+
+  constructor(
+    private route: ActivatedRoute,
+    private usersService: UsersService,
+  ) {}
+}
+```
+
+Finally the correct solution simply assigns user$ to be an observable which is created only once, template will have only one subscription and we will just react to updates in the data stream which happen on params change.
```
kristijanbambir:
```suggestion
Finally the correct solution simply assigns user$ to an observable which is created only once. The template will have only one subscription and we will just react to updates in the data stream which happen when params change.
```

PR #55, opened 2020-03-25
Feature/angular di infographic

filipjakov:
Added the [DI infographic](https://christiankohler.net/angular-dependency-injection-infographic) resource.

### DIFF
--- a/Angular/Angular guidelines and best practices.md
+++ b/Angular/Angular guidelines and best practices.md
@@ -1359,8 +1359,8 @@ class MyComponent implements OnInit, OnChanges {
 DI is utilized heavily in Angular, primarily for providing instances of either primitive values or class instances.
 
 DI is a very complex topic and requires a separate discussion which would be out of the scope of this handbook. But make sure to check out:
-- the [official DI guide](https://angular.io/guide/dependency-injection) which explains many details,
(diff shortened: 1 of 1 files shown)


Angular/Angular guidelines and best practices.md
```diff
@@ -1358,7 +1358,10 @@ class MyComponent implements OnInit, OnChanges {
 
 DI is utilized heavily in Angular, primarily for providing instances of either primitive values or class instances.
 
-DI is a very complex topic and requires a separate discussion which would be out of the scope of this handbook. Make sure to check out the [official DI guide](https://angular.io/guide/dependency-injection) which explains many details. You can also have a look at [this repo](https://github.com/fvoska/angular-di-demo) for some examples.
+DI is a very complex topic and requires a separate discussion which would be out of the scope of this handbook. But make sure to check out:
```
fvoska: I am sure @kristijanbambir will have a suggestion here - it is a bit strange to start a sentence with `But,`.

kristijanbambir: Correct. I propose the following:
```suggestion
DI is a very complex topic and requires a separate discussion which would be out of the scope of this handbook, but make sure you check out the following resources:
```

PR #57, opened 2020-04-21
Restructure react handbook

isBatak:
This is still WIP.
My plan here is to restructure files similar to Angular secretion and just rewrite examples to hooks to keep this PR as small as possible.

Some topics I want to cover in this PR:
- [ ] hooks
    - [ ] hooks lifecycle chart
- [ ] css-in-js
    - [ ] when animating on user interaction use style prop
    - [ ] ‘as’ prop, emotion ‘withComponent’, polymorphic component trade-offs and gotchas 
- [ ] tests
- [ ] project setup next.js
- [ ] i18next setup
- [ ] note about next.js examples folder
- [ ] next.js namespace folder naming inside components folder: ./src/components/reset-password/ComponentName/index.tsx (pages subfolder already has kebab-case  convention so let’s stick with that).
https://maxrozen.com/guidelines-improve-react-app-folder-structure/?ck_subscriber_id=887770662 this article has some good points
- [ ] boolean props prefix `is` and `has`


React/Getting started with React.md
```diff
@@ -0,0 +1,33 @@
+# Getting started with React
+
+### React
```
kristijanbambir, line 3: I think these headings should be larger, so `##` instead of `###`. The same goes for all of the other heading in this PR.

React/Getting started with React.md
```diff
@@ -1,12 +1,20 @@
 # Getting started with React
 
-### React
+## React
 
-- intro
+[React](https://reactjs.org/) is not a Framework, it is a JavaScript library for building user interfaces.
+with a declarative programming paradigm for building component based systems which could be reused anywhere.
```
kristijanbambir, line 6: Was this maybe meant to be part of the previous sentence?

React/Getting started with React.md
```diff
@@ -1,12 +1,20 @@
...
 
-### Next.js Framework
+In the next sections (no pun intended) we will cover some topics related to React ecosystem used inside our company.
```
kristijanbambir, line 8: What's the pun? 😄

React/Getting started with React.md
```diff
@@ -1,12 +1,20 @@
...
+
+## Next.js Framework
```
kristijanbambir, line 10: I would suggest to have Next.js as a separate chapter in the Handbook.

isBatak: Done!

React/Getting started with React.md
isBatak: `In the next sections` next is the name of the framework... I know I had better days :P

fvoska:
Can someone check if lector's comments from #37 still apply after the restructure?

React/Project structure.md
```diff
@@ -0,0 +1,361 @@
...
+Main goal of these files is to store functions and hooks that are **specific for that component**, so we could keep our root `hooks` and `utils` folders clean and for global usage purposes only.
+
+Example:
+
+```
+.
+├── ...
+└── AlbumsCarousel
+    ├── AlbumsCarousel.tsx
+    └── AlbumsCarousel.utils.ts
```
isBatak, line 130: Just `utils.ts` and `hooks.ts` without component name prefix.
Same reasons as for `elements.ts`

React/Project structure.md
```diff
@@ -0,0 +1,361 @@
...
+For example, your app could have an Admin dashboard on `/admin` route which has a fixed header and sidebars on both right and left, and scrollable content in the middle (you're most likely familiar with this kind of layout).
+
+You will define your different layouts inside `layouts` folder:
+
+```
+.
+.
+└── components
+    └── layouts
+        ├── admin-layout
```
isBatak, line 235: I think folder name should be PascalCase too
```
├── AdminLayout
     └── AdminLayout.tsx
```

isBatak: We could also have some `elements.ts` inside the folder

React/Project structure.md
```diff
@@ -0,0 +1,361 @@
...
+If you have a lot of style declarations inside you component file (enough to make the file difficult to read), you should create a separated file `elements.ts` where you will store you chakra factories.
+
+Example:
+
+```
+...
+.
+└── WelcomeCard
+    ├── WelcomeCard.tsx
+    └── WelcomeCard.elements.ts
```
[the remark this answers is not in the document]
DarkoKukovec: why is this an issue? @danijelbuhin and I tested this out and it worked without any issues.

React/Project structure.md
DarkoKukovec: why?

React/Project structure.md
DarkoKukovec: which folder? This is one of the reasons why I want to prefix elements with the "parent" component name, to avoid any standalone elements

React/Next/POC.md
```diff
@@ -0,0 +1,214 @@
...
+import { useRouter } from 'next/router';
+
+function isLoggedIn() {
+  const user = localStorage.getItem('user');
+  const userObject = JSON.parse(user);
+
+  return userObject ? !!userObject.token : false;
+}
+
+const withAuth = (Component) => (props) => {
```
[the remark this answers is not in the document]
DarkoKukovec: sure, but I don't really want to prolong the publishing any longer, so I would rather publish now, and update once we have it ready...

React/Project structure.md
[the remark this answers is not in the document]
isBatak: Really? 
I had an issue with webpack. 
But this works then great, maybe it was some setting on the underline project.

React/Next/POC.md
```diff
@@ -0,0 +1,214 @@
...
+
+```jsx
+// /components/higherOrderComponents/withModal.tsx
+import React from 'react';
+import { useRouter } from 'next/router';
+
+import useModalName from '../../customEffects/useModalName';
+
+const withModal = (Component) => (props) => {
+  const modalName = useModalName();
```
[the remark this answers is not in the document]
DarkoKukovec: sure, but same comment as above - let's merge what we have (unless it's completely wrong) and update later.

React/Project structure.md
isBatak: It's the same reason as for `elements.ts`, answered there ... it's great if this works as expected...

React/Project structure.md
isBatak:
```
├── admin-layout
     ├── AdminLayout.tsx
     └── AdminLayout.elements.tsx
 ```
 ```
├── AdminLayout
     ├── AdminLayout.tsx
     └── AdminLayout.elements.tsx
 ```
 
 The difference is in the folder name. This is just for consistency.
 - namespace folders `kebab-case`
 - component folders `PascalCase`

React/Next/POC.md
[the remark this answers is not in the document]
isBatak: Ok :D

PR #61, opened 2020-07-31
Styling section

veksi:


React Native/Styling.md
```diff
@@ -0,0 +1,211 @@
...
+      <AppCircle radius={50}>
+        {children}
+      </AppCircle>
+    )
+  }
+}
+```
+This means that when you're defining your styles inside `AppCircle`,  you're actually creating a normal `Circle` component, that has your styles attached to it. Important thing to mention here is that `AppCircle` inherits all the props from `Circle` component.
+
+When defining styled component in RN you can provide style properties directly:
```
DarkoKukovec, line 165: Maybe we should add when it's ok to use this? Seems practical for some dynamic styling, but we wouldn't want someone to just create an empty styled component and then add styles inline.

React Native/Styling.md
```diff
@@ -0,0 +1,215 @@
...
+
+Styling in RN can also be done with `styled components` which is more familiar approach for React developers.
+
+When starting a new project **`styled-components`** should be your default styling option as it consumes styling rules which are identical to React, and provides theming.
+
+Currently there are 2 main libraries used:
+1. [Styled components](https://styled-components.com/docs/basics#react-native)
+2. [Emotion](https://emotion.sh/docs/@emotion/native)
+
+There are almost no differences between them, but we decided that `styled-components` should be the option when starting a new project. Main reason for that is theming typings which can easily be defined when using `styled-components` while on the other hand `emotion` library doesn't have that clean solution yet.
```
RaulSusanjSamolov, line 75: I would rather suggest using RN styling in separate files for every component in combination with a single global file to define application themes (fonts, colors, frequently used styles). Styled-components contain 10 dependencies (about 1.15 MB in total) and must be tag evaluated, pre-processed, and then finally compiled into pure JavaScript bundles for the JS bridge thread to process it. Also, styled-components don't contain platform-specific styles and I'm also not sure if this type of styling in RN will be compatible with some 3rd party components (if used) because of the difference between web and mobile flex property. JUST A SUGGESTION :)

veksi: I was going through couple of permformance measure blog posts and overview of the same concern you have and basicly everyone had the same conclusion: if the elements are used smart way (for example using ListView instead of ScrollView for rendering large amount of items), styles don't have much impact on performance. Of course, styled-components do have some performance impact but it's not visible, atleast not in the applications we are writing now.
We may discuss if using this is ok in enterprise application with thousand or tens of thousand of items being rendered, but for now this will just bring React / web developers closer to RN.
But all in all I would like that we "define" some rules or maybe our own styleguide for native Stylesheet usage.
@nives113 I will create a sepearate task for this and assign it to me, so when we feel it's needed we can start working on it.

React Native/Styling.md
veksi: I've added simple explanation below code example.

React Native/Styling.md
RaulSusanjSamolov: Ok, this is fine then 😄

DarkoKukovec (approved):

PR #64, opened 2020-08-19
Add chapter about JS code quality

fvoska:


JavaScript/Code quality.md
```diff
@@ -0,0 +1,215 @@
...
+    "options": {
+      "parser": "angular"
+    }
+  }]
+}
+```
+
+If you just added Prettier to an existing codebase, you should probably run it once and let it format the whole codebase. This will probably create a lot of modifications which you can skim through and if it checks out you can commit the changes. It is possible that some things might break after Prettier is run. In particular, we noticed some issues with the way Prettier formats SCSS, so you might want to exclude SCSS from Prettier:
+
+```bash
```
kristijanbambir, line 103: I also had to tell VSCode not to format SCSS files when saving. I have the following the .vscode/settings.json:
```json
{
  "editor.formatOnSave": true,
  "[css]": {
    "editor.formatOnSave": false
  },
  "[scss]": {
    "editor.formatOnSave": false

  }
}
```

kristijanbambir (commented):
Nicely written! Perhaps there should be links to jump to configuration files faster if you've already read through the chapter, are starting a new project and need a quick way to set it up.

JavaScript/Code quality.md
fvoska: Did you have .prettierignore file with scss ignored?

PR #66, opened 2020-09-11
Custom Fonts

Hardronox:


React Native/Custom Fonts.md
```diff
@@ -0,0 +1,37 @@
...
+  },
+  assets: ["./assets/fonts/"], // stays the same
+};
+```
+
+- Link assets using react native link. Run in terminal:
+
+`react-native link`
+
+- Now you can use your font in your styles:
```
veksi, line 30: I would also explicitly say that when using custom fonts, css rule `font-weight` doesn't apply. To use different font `weight` we need to add separate font file with that weight, and then change the `font-familiy` in css to apply.

veksi (changes requested):

veksi (approved):

PR #75, opened 2020-12-02
react/chakra-ui

mamihelj09:


React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+
+### Installation:
+
+```
+npm i @chakra-ui/react @emotion/react @emotion/styled framer-motion
+```
+
+### Setup provider:
+
+For Chakra UI to work correctly, you need to setup the `ChakraProvider` at the root of your application.
```
fvoska, line 15: `setup` is a noun, use `set up` as a verb
```suggestion
For Chakra UI to work correctly, you need to set up the `ChakraProvider` at the root of your application.
```

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+<Box sx={{ width: '400px'}}>Lorem Ipsum is simply dummy text</Box>
+```
+
+For more about style props read [docs](https://chakra-ui.com/docs/features/style-props)
+
+#### The `as` prop
+
+The `as` prop is a feature that all Chakra UI components have and it allows you to pass an HTML tag or component to be rendered.
+
+For example, say you are using a Button component, and you need to make it a link instead. You can compose a and Button like this:
```
fvoska, line 58: To make it clearer that `a` is not the article from English language (a/the), but the `anchor` element:

```suggestion
For example, say you are using a `Button` component, and you need to make it a link instead. You can compose `a` and `Button` like this:
```

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+
+```tsx
+<Button as="a" variant="outline" href="https://chakra-ui.com">
+  Lorem Ipsum
+</Button>
+```
+
+This allows you to use all of the Button props and all of the a props without having to wrap the Button in an a component.
```
fvoska, line 66:
```suggestion
This allows you to use all of the `Button` props and all of the `a` props without having to wrap the `Button` in an `a` component.
```

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+
+In case we need to skip cretan breakpoint, `null` is passed at that position in the array to avoid generating unnecessary CSS.
+
+Example:
+
+```jsx
+<Text fontSize={["24px", null, "56px"]}>Lorem Ipsum is simply dummy text</Text>
+```
+
+Array and Object syntax work for every style prop in the them specification, which means you can change most properties at a given breakpoint.
```
fvoska, line 250: What did you mean by `in the them specification`?

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+
+- `baseStyles`
+- `sizes`
+- `variants`
+- `defaultProps`
+
+Base style are styles that all button types share.
+
+```tsx
+const baseStyle = {
```
fvoska, line 269: How would you ensure that there are no typos in this object? For example: `lineHeigth` instead of `lineHeight`? Is there a type for this object?

Also how would you ensure that only valid values are used? For example: `cursor: "foo"`

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+    border: '1px solid',
+    borderColor: 'primary.100',
+    _hover: {
+      bg: 'primary.100',
+    }
+  },
+};
+```
+
+Sizes determent all the component sizes like `width`, `height`, `font-size` etc
```
fvoska, line 313: Did you mean `determine`?

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+```
+
+Naming used for sizes:
+
+- `lg`
+- `md`
+- `sm`
+- `xs`
+
+Default values are default values for `size` and `variant`
```
fvoska, line 339: This can be a bit confusing to read, maybe change it to this?
```suggestion
`defaultProps` are default values for `size` and `variant`
```

Similar comment for `Sizes determent all` as well, maybe change it to
```
`sizes` determine....
```

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+      styles={selectStyle}
+      {...props}
+    />
+  );
+};
+```
+
+Each key added to the theme needs to be added to `selectStyle`.
+
+Additionally custom components can be added to react-select like `SelectContainer`
```
fvoska, line 477:
```suggestion
Additionally, custom components can be added to react-select like `SelectContainer`
```

fvoska (changes requested):

React/CSS-in-JS/Chakra.md
mamihelj09: theme*

React/CSS-in-JS/Chakra.md
mamihelj09: > How would you ensure that there are no typos in this object? For example: lineHeigth instead of lineHeight? Is there a type for this object?
I added `StyleObjectOrFn` type for that

> Also how would you ensure that only valid values are used? For example: cursor: "foo"
This is a problem because chakra doesn't have a type for that and it would be hard to create one because the value can be some css value or something from a theme or array (cc @isBatak)

fvoska (approved):

safo6m (approved):

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+}
+
+export default App;
+```
+
+Note:
+
+- For Next.js, you need to set this up in `pages/_app.tsx`
+- For Create React App, you need to set this up in `index.tsx`
+
```
isBatak, line 37: Add a note that ChakraProvider uses emotion ThemeProvider under the hood. 
https://github.com/chakra-ui/chakra-ui/blob/develop/packages/react/src/chakra-provider.tsx#L57
https://github.com/chakra-ui/chakra-ui/blob/develop/packages/system/src/providers.tsx#L12

Mention that it uses [styled-system](https://styled-system.com/) too
https://github.com/chakra-ui/chakra-ui/blob/develop/packages/styled-system/src/system.ts#L1
Explain [responsive-styles](https://styled-system.com/responsive-styles) props 

Pseudo props should also be mentioned somewhere https://chakra-ui.com/docs/features/style-props#pseudo, this is specific to Chakra UI

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+// shorthand width prop
+<Box w="400px">Lorem Ipsum is simply dummy text</Box>
+
+// sx prop
+<Box sx={{ width: '400px'}}>Lorem Ipsum is simply dummy text</Box>
+```
+
+For more about style props read [docs](https://chakra-ui.com/docs/features/style-props)
+
+#### The `as` prop
```
isBatak, line 54: This is called `polymorphic prop` and it's actually a feature of Emotion https://emotion.sh/docs/styled#as-prop borrowed from styled-components https://styled-components.com/docs/api#as-polymorphic-prop.
You should mention it in this section.

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+
+For example, say you are using a `Button` component, and you need to make it a link instead. You can compose `a` and `Button` like this:
+
+```tsx
+<Button as="a" variant="outline" href="https://chakra-ui.com">
+  Lorem Ipsum
+</Button>
+```
+
+This allows you to use all of the `Button` props and all of the `a` props without having to wrap the `Button` in an `a` component.
```
isBatak, line 66: You can mention that this also work `<Button as={Link} someLinkProp={value} />`
- Link is another component
- Link props will become available on Button
- styles will be combined

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+  },
+};
+```
+
+Then in `props` we have access button props, `theme` object and `colorMode`.
+
+Note:
+
+- `mode(lightMode, darkMode)(props)` function is the same as `colorMode === "dark" ? darkMode : lightMode`.
+- `mode` function is part of `chakra-ui/theme-tools` package.
```
isBatak, line 132: there is also `orient` https://github.com/chakra-ui/chakra-ui/blob/develop/packages/theme-tools/src/component.ts#L44 
Example: https://github.com/chakra-ui/chakra-ui/blob/e20efc1d8d42ce9d3c2280a5731d59e771de53c4/packages/theme/src/components/slider.ts#L8

And a lot of color transformation helpers https://github.com/chakra-ui/chakra-ui/blob/develop/packages/theme-tools/src/color.ts
`transparentize` is a cool one
Example: https://github.com/chakra-ui/chakra-ui/blob/abac2f999f3996ffc3ce575de07126dc9cc41bbb/packages/theme/src/components/badge.ts#L13

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+
+Example:
+
+```jsx
+<Text fontSize={["24px", null, "56px"]}>Lorem Ipsum is simply dummy text</Text>
+```
+
+Array and Object syntax work for every style prop in the theme specification, which means you can change most properties at a given breakpoint.
+
+To create custom breakpoints read [docs](https://chakra-ui.com/docs/features/responsive-styles#customizing-breakpoints).
```
isBatak, line 252: Maybe describe `orient` here also.

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,498 @@
...
+```
+
+#### useColorModeValue
+
+`useColorModeValue` is React hook that takes 2 arguments, first is value for light mode and second is value for dark mode, and returnees the the value based on the active color mode
+
+```tsx
+const value = useColorModeValue(lightModeValue, darkModeValue);
+```
+
```
isBatak, line 380: Maybe we could add one more section dedicated to the chakra factory https://chakra-ui.com/docs/features/chakra-factory.
It could be used similar to emotion `styled` component but it will be aware of the theme props.
```
const Card = chakra("div", {
  baseStyle: {
    shadow: "lg",
    rounded: "lg",
    bg: "white",
  },
})
```

isBatak (changes requested):

React/CSS-in-JS/Chakra.md
mamihelj09: I added chakra factory section + example with chakra factory function and `react-datepicker`, similar to there's `react-input-autoresize` [example](https://chakra-ui.com/docs/features/chakra-factory#chakra-factory-function)

React/CSS-in-JS/Chakra.md
```diff
@@ -55,15 +58,35 @@ For more about style props read [docs](https://chakra-ui.com/docs/features/style
```
isBatak, line 58: The `as` polymorphic prop

React/CSS-in-JS/Chakra.md
```diff
@@ -251,6 +327,59 @@ Array and Object syntax work for every style prop in the theme specification, wh
...
+    shadow: "lg",
+    rounded: "lg",
+    bg: "white",
+  },
+});
+```
+
+This reduces the need to create custom component wrappers and name them. Syntax is available for common html elements. See the reference for the full [list of elements](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/system/src/system.utils.ts#L9) supported.
+
+Chakra factory function can also that convert jsx element to chakra-enabled components so you can pass style props to them.
```
isBatak: This is not clear enough. I don't understand what you want to say

React/CSS-in-JS/Chakra.md
```diff
@@ -251,6 +327,59 @@ Array and Object syntax work for every style prop in the theme specification, wh
...
+For example `react-datepicker` can be wrapped in chakra factory function to be possible to pass style props to `<DatePicker />`
+
+Example:
+
+```tsx
+import DatePicker from "react-datepicker";
+import { chakra } from "@chakra-ui/react";
+
+export const Datepicker = () => {
+  const ChakraDatepicker = chakra(DatePicker);
```
isBatak, line 371: this should be moved outside the founction

React/CSS-in-JS/Chakra.md
```diff
@@ -251,6 +327,59 @@ Array and Object syntax work for every style prop in the theme specification, wh
...
+
+  return <ChakraDatepicker sx={{ width: "100%", bg: "blue.100" }} />;
+};
+```
+
+Chakra factory function will pass `class` to `DatePicker input` component with all styles. In this case `input` will be rendered with custom `bg` and `width` styles.
```
isBatak, line 377: `class` => `className`

React/CSS-in-JS/Chakra.md
```diff
@@ -251,6 +327,59 @@ Array and Object syntax work for every style prop in the theme specification, wh
...
+    bg: "white",
+  },
+});
+```
+
+This reduces the need to create custom component wrappers and name them. Syntax is available for common html elements. See the reference for the full [list of elements](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/system/src/system.utils.ts#L9) supported.
+
+Chakra factory function can also that convert jsx element to chakra-enabled components so you can pass style props to them.
+
+For example `react-datepicker` can be wrapped in chakra factory function to be possible to pass style props to `<DatePicker />`
```
isBatak, line 362: `to be possible` sounds weird... but I'm not an English expert :D

isBatak (changes requested):

React/CSS-in-JS/Chakra.md
mamihelj09: I removed this because we have something similar few lines above.

React/CSS-in-JS/Chakra.md
```diff
@@ -359,22 +359,22 @@ chakra("button", {
 
 This reduces the need to create custom component wrappers and name them. Syntax is available for common html elements. See the reference for the full [list of elements](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/system/src/system.utils.ts#L9) supported.
 
-For example `react-datepicker` can be wrapped in chakra factory function to be possible to pass style props to `<DatePicker />`
+For example `react-datepicker` can be wrapped in chakra factory function so we can pass `sx` style props to `<DatePicker />`
```
isBatak, line 362: `sx` **or** style props

isBatak (approved):

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,629 @@
...
+
+For example `react-datepicker` can be wrapped in chakra factory function so we can pass `sx` or style props to `<DatePicker />`
+
+Example:
+
+```tsx
+import DatePicker from "react-datepicker";
+import { chakra } from "@chakra-ui/react";
+
+const ChakraDatepicker = chakra(DatePicker);
```
isBatak, line 370: Can we use `Styled` prefix... Chakra use it in the source code... so
const StyledaDatepicker = chakra(DatePicker);

isBatak: For example here https://github.com/chakra-ui/chakra-ui/blob/develop/packages/menu/src/menu.tsx#L71

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,629 @@
...
+```
+
+Note:
+
+- For Next.js, you need to set this up in `pages/_app.tsx`
+- For Create React App, you need to set this up in `index.tsx`
+
+### Style props
+
+Style props are a way to alter the style of a component by simply passing props to it. It helps to save time by providing helpful shorthand ways to style components.
```
kristian240, line 43: I guess this would be preffered way of doing it?

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,629 @@
...
+### Global styles
+
+Global styles are theme-aware styles you can apply to any html element globally.
+
+They are defined in `src/styles/theme/styles.ts` file.
+
+<!-- TODO maybe what should be defined in global -->
+
+Example:
+
```
kristian240, line 260: Can we also mention `resetCss` prop on ChakraProvider?

https://chakra-ui.com/docs/migration#2-update-the-themeprovider

mamihelj09: I will add it

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,629 @@
...
+Note:
+
+- For Next.js, you need to set this up in `pages/_app.tsx`
+- For Create React App, you need to set this up in `index.tsx`
+
+### Style props
+
+Style props are a way to alter the style of a component by simply passing props to it. It helps to save time by providing helpful shorthand ways to style components.
+
+Alternative to that is using sx prop that takes an object with style rules.
```
DarkoKukovec, line 49: It's fine to mention alternatives, but the preferred way of using it should be clear, together with some examples when the alternative might be better.

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,629 @@
...
+<Box sx={{ width: '400px'}}>Lorem Ipsum is simply dummy text</Box>
+```
+
+For more about style props read [docs](https://chakra-ui.com/docs/features/style-props)
+
+#### The `as` polymorphic prop
+
+The `as` prop is a feature that all Chakra UI components have and it allows you to pass an HTML tag or component to be rendered.
+
+`as` prop (`polymorphic prop`) is feature from emotion borrowed from styled-components
```
DarkoKukovec, line 65:
```suggestion
`as` prop (`polymorphic prop`) is a feature of emotion borrowed from styled-components
```

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,629 @@
...
+```
+
+```tsx
+// object syntax
+<Text fontSize={{ base: "24px", md: "40px", lg: "56px" }}>
+  Lorem Ipsum is simply dummy text
+</Text>
+```
+
+In case we need to skip cretan breakpoint, `null` is passed at that position in the array to avoid generating unnecessary CSS.
```
DarkoKukovec, line 322:
```suggestion
In case we need to skip a certain breakpoint, `null` is passed at that position in the array to avoid generating unnecessary CSS.
```

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,629 @@
...
+
+Example:
+
+```jsx
+<Text fontSize={["24px", null, "56px"]}>Lorem Ipsum is simply dummy text</Text>
+```
+
+Array and Object syntax work for every style prop in the theme specification, which means you can change most properties at a given breakpoint.
```
DarkoKukovec, line 330: Again, we should define which one is the preferred way of using it.

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,629 @@
...
+    cursor: "not-allowed",
+    boxShadow: "none",
+  },
+  _hover: {
+    cursor: "pointer",
+  },
+};
+```
+
+<!-- TODO change this example to use color schema -->
```
DarkoKukovec, line 417: um?

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,629 @@
...
+    bg: 'primary.100',,
+    color: 'white',
+    _hover: {
+      bg: 'primary.400',
+    },
+    _disabled: {
+      bg: 'neutral.200',
+    }
+  },
+	outline: (props: Dict): SystemStyleObject => ({
```
DarkoKukovec, line 433: Seems like you mixed spaces and tabs here :)

doroz0:
I think we should use the `sx` prop very sparingly.  
From what I saw looking at few ThemeUI projects, where `sx` prop exists, is that we use it way too much and it makes the code hard to read or understand. If a component needs extra styling but a custom variant is not necessary, let's not use the `sx` prop, and let's create a separate component. *(but one-offs are ok, see the first* `BoxItem` *in my example)*  
If the component isn't "big" enough to be a separate file, I would advise we use Chakra Factory and define the component outside of the render function. The result is a cleaner code.

Basically what I'm trying to say is that I'd rather have this:

```jsx
import { Stack, Box } from '@chakra-ui/react';

const BoxItem = chakra(Box, {
    baseStyle: {
        shadow: "lg",
        rounded: "lg",
        bg: "white",
    }
})

const ListOfBoxItems: FC = () => (
    <Stack spacing="16px">
        <BoxItem sx={{ bg: "blue.100" }}>...</BoxItem>
        <BoxItem>...</BoxItem>
    </Stack>
);
```
  
than this:
  
```jsx
import { Stack, Box } from '@chakra-ui/react';

const ListOfBoxItems: FC = () => (
    <Stack spacing="16px">
        <Box
            sx={{
                shadow: "lg",
                rounded: "lg",
                bg: "blue.100",
            }}
        >
            ...
        </Box>
        <Box
            sx={{
                shadow: "lg",
                rounded: "lg",
                bg: "white",
            }}
        >
            ...
        </Box>
    </Stack>
);
```
  
Naming is also something that needs discussing here. I think it's better to name components by their purpose instead of what they are. In this case, someone would name the `BoxItem` as `StyledBox`. To a reader, this means the `Box` component has been modified but they can't know how or why unless they look at the styling.  
Here's a small example that probably isn't the best but it's enough to show what I'm thinking.  
Let's say we're writing an album component that has two thumbnails and some extra details per thumbnail *(just a reason so a wrapper/container around the image is needed)*.  
  
I'd rather have this:
  
```jsx
const ThumbnailContainer = chakra(Box, {
    baseStyle: {
        shadow: "lg",
        rounded: "lg",
    }
})

const Album: FC = () => (
    <HStack spacing="4px">
        <ThumbnailContainer>...</ThumbnailContainer>
        <ThumbnailContainer>...</ThumbnailContainer>
    </HStack>
);
```
  
than this:
  
```jsx
const StyledBox = chakra(Box, {
    baseStyle: {
        shadow: "lg",
        rounded: "lg",
    }
})

const Album: FC = () => (
    <HStack spacing="4px">
        <StyledBox>...</StyledBox>
        <StyledBox>...</StyledBox>
    </HStack>
);
```

React/CSS-in-JS/Chakra.md
isBatak: Style props for one-off styles
```
<Button display="block" mt="md" />
```

`sx` prop when you have to merge multiple objects
```
const ExoticComponent = (props) => {
  const styles = useStyleConfig('ExoticComponent', props);
 
  const defaultStyles = {
   display: 'block',
  };

  return <chakra.div sx={{ ...styles, ...defaultStyles }} />;
}
```
You have control over specificity and merge order. 
Also, there is `__css` which will get merged first, before the `sx`, so I guess we should use `__css` in this case.
https://github.com/chakra-ui/chakra-ui/blob/085891be806b855af90b86367f5b26e8151c3ff5/packages/system/src/system.ts#L106

```
/**
 * Style resolver function that manages how style props are merged
 * in combination with other possible ways of defining styles.
 *
 * For example, take a component defined this way:
 * ```jsx
 * <Box fontSize="24px" sx={{ fontSize: "40px" }}></Box>
 * ```
 *
 * We want to manage the priority of the styles properly to prevent unwanted
 * behaviors. Right now, the `sx` prop has the highest priority so the resolved
 * fontSize will be `40px`
 */
 ```

**Comment by mamihelj09** [comment]
@doroz0 I agree whit ur comment.
We will add a section defining when to use `sx` and when not based on ur example. Also we will add a section about `__css`.

Regarding the naming of components, I agree we should name them based on their purpose but I'm not sure if we should define this here in Chakra section of handbook. Maybe a better place for that would be in project structure section of handbook. (cc @isBatak )

**Comment by DarkoKukovec** [comment]
@doroz0 @isBatak any more comments or can this be merged?

File: React/CSS-in-JS/Chakra.md [hunk 30]
```diff
@@ -0,0 +1,657 @@
...
+    bg: "white",
+  }}
+>
+  Click me
+</chakra.button>;
+```
+
+This reduces the need to create custom component wrappers and name them. Syntax is available for common html elements. See the reference for the full [list of elements](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/system/src/system.utils.ts#L9) supported.
+
+Chakra factory function should be only used in atom components. They will be defined with base style and can be overridden by `sx` prop because `baseStyle` is merged before `sx` ([merge order](https://github.com/chakra-ui/chakra-ui/blob/085891be806b855af90b86367f5b26e8151c3ff5/packages/system/src/system.ts#L104))
```
doroz0, line 124: I don't think we should limit chakra factory to be used only in atom components. They should be allowed (and preferred) everywhere except pages (as we discussed).

Components created using chakra factory can be overridden by also using chakra factory (which would in my opinion also be the preferred way), so let's mention that here.

```
const Button = chakra("button", {...});
const FbButton = chakra(Button, {...});
```

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,657 @@
...
+    shadow: "lg",
+    rounded: "lg",
+    bg: "white",
+  }}
+>
+  Click me
+</chakra.button>;
+```
+
+This reduces the need to create custom component wrappers and name them. Syntax is available for common html elements. See the reference for the full [list of elements](https://github.com/chakra-ui/chakra-ui/blob/develop/packages/system/src/system.utils.ts#L9) supported.
```
doroz0, line 122: The list just contains elements that you can put in as a string.
You can easily pass any element to chakra:

```
chakra((props) => <someelement {...props} />, {...})
```

Or since `className` is how chakra styles components:

```
chakra(({className}) => <someelement {...{className}} />, {...})
```

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,657 @@
...
+
+Example:
+
+```tsx
+import DatePicker from "react-datepicker";
+import { chakra } from "@chakra-ui/react";
+
+const StyledaDatepicker = chakra(DatePicker);
+
+export const Datepicker = () => <StyledaDatepicker w="100%" bg="blue.100" />;
```
doroz0, line 138: Well, you're already using the chakra factory so might as well style it there.

```
const Datepicker = chakra(DatePicker, {
  baseStyle: {
    w: '100%', // or 'full'
    bg: 'blue.100',
  },
});
```

isBatak: It seems this demonstrates how to use theme-aware one-off styles with components that are not part of chakra.
A SImilar thing could be achieved with polymorphic `as` prop.
`<Box as="DatePicker" w="100%" bg="blue.100" />`
This is totally fine to me...

DarkoKukovec:
@mamihelj09 can you please go trough the latest feedback? We should merge this PR and then we can do some other tweaks later if needed.

React/CSS-in-JS/Chakra.md
```diff
@@ -0,0 +1,657 @@
...
+
+Example:
+
+```jsx
+<Text fontSize={["24px", null, "56px"]}>Lorem Ipsum is simply dummy text</Text>
+```
+
+Array and Object syntax work for every style prop in the theme specification, which means you can change most properties at a given breakpoint. Preferred way is to use array.
+
+To create custom breakpoints read [docs](https://chakra-ui.com/docs/features/responsive-styles#customizing-breakpoints).
```
kristian240, line 407: I had some issue with this while our project `tsconfig` had `strict: false`. We should add a note that in order for everything to work as expected in `chakra-ui@1.3.2` and later you need to have `strict: true` in your `tsconfig` otherwise you will get a type error that `breakpoint.toString` are incompatible.

Here is the sandbox of the error: https://codesandbox.io/s/sleepy-sanne-t3fk4

DarkoKukovec (approved):

doroz0 (approved):

PR #60, opened 2020-07-17
Added testing section in Frontend category

mislavmiocevic:
This write up is about testing in general and some guidelines for Frontend testing.

It includes:
- what is testing and why is it important
- types of testing
- short introduction to TDD
- some notes about testing in team

PR #79, opened 2020-12-23
Project structure

danijelbuhin:
Currently covered:
 - Components organization (I kept focus only on the atomic structure since as far as I could tell, this is something we are leaning towards to)
 - Tests
 - Store
 - Fetchers

Inspiration:
- NLBL, soundpark, TPE, react-example and pd-vault repos
- [Atomic Design With React And Bit: Simplify a Complex UI](https://dev.to/giteden/atomic-design-with-react-and-bit-simplify-a-complex-ui-1k1f)
- [Atomic design methodology](https://atomicdesign.bradfrost.com/)

I did not wanted to go into very deep explanations to avoid a lot of text. Also, I did not cover some most common stuff like:
- Assets
- Locales
- Typings (interfaces, enums..)

If you think these points should be included, I could make section like `Misc`.

### DIFF
--- a/React/NextJs/ProjectStructure.md
+++ b/React/NextJs/ProjectStructure.md
@@ -1,10 +1,6 @@
 # Project file structure
 
 ## Organizing components
-(For PR Review):
-I believe this section will be refactored once we get around with all chakra stuff and component organization methodology.
-I kept going with only atomic design approach.
-
 ### UI Components
 
 Before implementing [atomic design methodology](https://github.com/danilowoz/react-atomic-design), our common practice was separating components by the scope of the page + core components + shared components like this:
@@ -16,16 +12,16 @@ Lowercase folders were indicators of a page scope (except of core).
 ├── src
 └── components
     ├── core
-    │   └── Button
-    │       ├── index.tsx
+    │   └── button
+    │       ├── Button.tsx
     │       └── stories.tsx
     ├── admin
-    │   └── WelcomeBanner
-    │       └── index.tsx
-    ├── Carousel
-    │   └── index.tsx
-    └── Card
-        └── index.tsx
+    │   └── welcome-banner
+    │       └── WelcomeBanner.tsx
+    ├── carousel
+    │   └── Carousel.tsx
+    └── card
+        └── Card.tsx
 ```
 
 With atomic design, we would completely separate all UI components and build them as a blocks of atoms, molecules and organisms (together with templates and pages for putting everything together at the end) 
@@ -36,17 +32,17 @@ src
 └── components
     └── ui
         ├── atoms
-        │   └── Button
+        │   └── button
         │       ├── Button.tsx
         │       └── stories.tsx
         ├── molecules
-        │   ├── Card
-        │   │   └── index.tsx
-        │   └── WelcomeBanner
-        │       └── index.tsx
+        │   ├── card
+        │   │   └── Card.tsx
+        │   └── welcome-banner
+        │       └── WelcomeBanner.tsx
         └── organisms
-            └── Carousel
-                └── index.tsx
+            └── carousel
+                └── Carousel.tsx
 ```
 
 ## Different molecule layouts
@@ -58,11 +54,11 @@ In this case, inside a specific molecule, we could add a subfolder `layouts` (no
 ```
 .
 └── molecules
-    └── Card
+    └── card
         ├── layouts
         │   ├── Desktop.tsx
         │   └── Mobile.tsx
-        └── index.tsx
+        └── Card.tsx
 ```
 
 For this case, inside `index.tsx` we would have something like this:
@@ -95,12 +91,12 @@ You will define your different layouts inside `layouts` folder:
 .
 └── components
     └── layouts
-        ├── AdminLayout
-        │   └── index.tsx
-        ├── MainLayout
-        │   └── index.tsx
-        └── BlogLayout
-            └── index.tsx
+        ├── admin-layout
+        │   └── AdminLayout.tsx
+        ├── main-layout
+        │   └── MainLayout.tsx
+        └── blog-layout
+            └── BlogLayout.tsx
 ```
 
 Then, when creating your routes (pages), you will wrap your page in the layout that represents the current page:
@@ -128,14 +124,14 @@ export default function Admin() {
 
 ## Setting up store
 
-Your datx store will be placed in the root of the `src` folder as follows:
+Your datx store and models will be placed in the root of the `src` folder as follows:
 
 ```
 src
+├── models
+│   ├── User.ts
+│   └── Session.ts
 └── store
-    ├── models
-    │   ├── User.ts
-    │   └── Session.ts
     ├── utlis
     │   ├── config.ts
     │   └── initializeStore.ts
@@ -150,7 +146,7 @@ In `src/fetchers` you will organize your (swr) fetchers, usually by the model on
 // src/fetchers/user.ts
 
 import { AppCollection } from '../store';
-import { User } from '../store/models/User';
+import { User } from '../models/User';
 
 export async function fetchUser(store: AppCollection, id: string): Promise<User> {
   try {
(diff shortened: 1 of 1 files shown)


React/NextJs/ProjectStructure.md
```diff
@@ -0,0 +1,212 @@
...
+    │   └── Session.ts
+    ├── utlis
+    │   ├── config.ts
+    │   └── initializeStore.ts
+    └── index.ts
+```
+
+## Fetchers
+
+In `src/fetchers` you will organize your (swr) fetchers, usually by the model on which you will make API calls.
```
fvoska, line 147: Not a big deal, but a bit strange folder hierarchy with models and fetchers.

Models are in `src/store/models` and fetchers that are in  `src/fetchers` are using those models. I would expect that the models and fetchers dirs are siblings (on the same level) instead of models being in a subdir.

Would one of the following options make sense?

1.
```
src/
  - store/
  - models/
  - fetchers/
```
2.
```
src/
  - store/
    - models/
    - fetchers/
```

React/NextJs/ProjectStructure.md
```diff
@@ -0,0 +1,212 @@
...
+        └── index.tsx
+```
+
+With atomic design, we would completely separate all UI components and build them as a blocks of atoms, molecules and organisms (together with templates and pages for putting everything together at the end) 
+
+```
+src
+.
+└── components
+    └── ui
```
fvoska, line 37: Are ALL atoms, molecules and organisms under top-level `components/ui` dir, or do you allow for multiple `ui` dirs for multiple features? For example, some page might have some organism component that is used only on that page. It wouldn't make sense to place that organism under global `components/ui`, I would place it in some dir like `page/my-page/components/ui/organism`. Basically allow to have this structure that you explained here both globally and for feature-specific cases.

The reason why I am saying this is that if an app is really large, you will have 100+ organism in `components/ui/organism` and some of them might be used only on certain pages/features and some are real globals.

React/NextJs/ProjectStructure.md
```diff
@@ -0,0 +1,212 @@
...
+
+
+## Setting up theming and styles
+
+_Sync with @mamihi_official_
+* Styles will probably be a part of components organization
+
+## Tests
+
+(For PR Review):
```
fvoska, line 179: I would also add info about folder structure for tests for other units as well, like store, fetchers, etc.

We should also write some "recipes" about actually writing these tests for various types of units (components, hooks, store, fetchers, etc.). But those details are some other markdown file, perhaps `React/NextJs/Testing.md, this file is just for structure.

fvoska (approved):

fvoska (changes requested):
Changed from Approved to Request Changes

React/NextJs/ProjectStructure.md
DarkoKukovec: I wouldn't mind if the models were in src - they are quite important and nesting them inside of the store folder is basically just mirroring a datx implementation detail, which is not necessary.

React/NextJs/ProjectStructure.md
DarkoKukovec: I agree, but this is maybe a topic for a different section/PR. It should be somewhere in between this one and https://github.com/infinum/frontend-handbook/pull/60

React/NextJs/ProjectStructure.md
DarkoKukovec: @isBatak should probably give his feedback on this, but I believe we talked about the possibility of nesting components in such cases?
It would be good to explicitly mention it here.

React/NextJs/ProjectStructure.md
```diff
@@ -0,0 +1,212 @@
+# Project file structure
+
+## Organizing components
+(For PR Review):
+I believe this section will be refactored once we get around with all chakra stuff and component organization methodology.
+I kept going with only atomic design approach.
+
+### UI Components
+
+Before implementing [atomic design methodology](https://github.com/danilowoz/react-atomic-design), our common practice was separating components by the scope of the page + core components + shared components like this:
```
DarkoKukovec, line 10: I'm not sure this has any value in the handbook since it will be obsolete and forgotten. Maybe it would be more useful as a comparison once the team starts using it.

React/NextJs/ProjectStructure.md
```diff
@@ -0,0 +1,212 @@
...
+  │   │   ├── useModal.test.ts
+  │   │   └── useNotification.test.ts
+  │   ├── useModal.ts
+  │   └── useNotification.ts
+  └── utils
+      ├── __tests__
+      │   └── generateFileTree.test.ts
+      └── generateFileTree.ts
+
+```
```
DarkoKukovec, line 212: Maybe a recap would be useful here, as a more complex example that includes most of the covered cases?

React/NextJs/ProjectStructure.md
adrianbicanic: Fetchers might not be related to the store (for example, if you are using `swr` to fetch google places), so I like the first, simplest option better. Also, it always bothered me that models are inside the store folder, although they are used globally (their purpose is not only for using them with the store, but also to get autocomplete inside components, utils, etc.).

React/NextJs/ProjectStructure.md
adrianbicanic: @DarkoKukovec Was there any idea of how to specifically nest such components, since they are not related to a context? 
@fvoska's example (`page/my-page/components/ui/organism`) also put's it into a context, which kind of beats the purpose ("At this level, the components begin to have the final shape, but they are still ensured to be independent, portable and reusable enough to be reusable in any content."), but in reality projects always get clogged and hard to read with a lot of files in a flat structure.

danijelbuhin:
Updates (synced with @mamihelj09): 
- Fixed component folder names
- Updated testing and styling sections

@isBatak and I will have a short meeting tomorrow about the extending atomic structure and overall progress of this handbook section, so more updates will soon come.

React/NextJs/ProjectStructure.md
filipjakov: @fvoska as quoted [here](https://github.com/infinum/JS-React-Example/blob/chakra/docs/atomic-design.md#organisms):

> Organisms are the combination of molecules that work together or even with atoms that compose more elaborate interfaces. At this level, the components begin to have the final shape, **but they are still ensured to be independent portable, and reusable enough to be reusable in any content**.

With that being said, organisms should still be reusable across different contexts/pages, and I believe their place is in the global ui folder.

I believe the mindset of context-bound (per page) organisms comes from one (or more) of the following situations:
- small project === small number of pages/features which results in a small number or organisms (no problem in traversing them at this point)
- component architecture is not setup as per the methodology violating the beforementioned principle (bolded in the quote)
- the design system is either bad and/or "versatile" resulting in the "inherent" creation of independent components. This might be out of scope here, but as projects get bigger there should be a consensus with the designers yadayada

Worst case scenario: you have a huge amount of code and you would like to scope the organisms in some manner; I'd propose to create subfolders which are scoped to either features or pages like so:

```
ui/
  organisms/
    feature1/
      some-organism
    feature2/
      some-organism
```

isBatak: @fvoska  Per page component namespacing is kinda a wrong mindset here.
Organisms are basically "container" components but in the new React ecosystem, you don't have to fetch data at the top-level anymore, instead, you can fetch data in any component and still be able to reuse that component wherever you want. We collocate data fetching with a component to make it highly reusable, therefore you don't have to model your component based on the place where it is rendered, instead, you can model it by the data you are presenting.

This is the definition of the Organism that I found on the internet  
> Organisms are the combination of molecules that work together or even with atoms that compose more elaborate interfaces. At this level, the components begin to have the final shape, but they are still ensured to be independent, portable, and reusable enough to be reusable in any content.

I copied the whole article here, maybe we can reuse some definitions for our handbook.
https://github.com/infinum/JS-React-Example/blob/chakra/docs/atomic-design.md#organisms

[Templates](https://github.com/infinum/JS-React-Example/blob/chakra/docs/atomic-design.md#templates) and a good design system (like Chakra UI), kinda resolve this problem and need to scope components to a specific page. They shape the organisms to fit on a specific page... 

In conclusion, all of this is really new to us and we need to take it slow. 
I wouldn't go with per page namespacing because it seems like a wrong abstraction.

React/NextJs/ProjectStructure.md
```diff
@@ -0,0 +1,224 @@
...
+Sometimes, you might want to create a molecule specific for mobile and desktop. You might use a tool like [fresnel](https://github.com/artsy/fresnel) for detecting mediaqueries.
+
+In this case, inside a specific molecule, we could add a subfolder `layouts` (not to be confused with actual layout described below) to define how our molecule would look like on specific media query.
+
+```
+.
+└── molecules
+    └── card
+        ├── layouts
+        │   ├── Desktop.tsx
```
isBatak, line 59: Can we name it like this
`Card.desktop.tsx`, `Card.mobile.tsx`, `Card.tablet.tsx`
it is easier to search for a component.
And react by itself will introduce naming like this with server-side components, e.g. `Card.server.tsx` and `Card.client.tsx`

React/NextJs/ProjectStructure.md
```diff
@@ -0,0 +1,224 @@
...
+
+```
+src
+.
+└── components
+    └── ui
+        ├── atoms
+        │   └── button
+        │       ├── Button.tsx
+        │       └── stories.tsx
```
isBatak, line 37: We should name this `Button.stories.tsx` because it is easier to search through the project

React/NextJs/ProjectStructure.md
isBatak: > React had a vision that goes beyond the browser. they understood that a clean component model is the unifying factor, that splitting up the view layer is a logical and technical mistake. the view and the code that forms it belong together, only state should be separate.

danijelbuhin:
Latest update:

- Removed atomic design structure for components
- Changed the order of some sections to keep the reading flow meaningful

React/NextJs/ProjectStructure.md
```diff
@@ -30,64 +30,60 @@ src
...
-        └── index.tsx
+├── shared
+│   └── WelcomeBox
+│       └── WelcomeBox.tsx
+├── admin
+│   └── AdminBanner
+│       └── AdminBanner.tsx
+└── Gallery
+    ├── Gallery.tsx
+    └── components
```
DarkoKukovec, line 57: Hmm, how would the nesting work here? Would we allow nesting of such components or would it only be on an "elements file" basis?

DarkoKukovec: I'm not necessary against it, but it should be explicitly defined

DarkoKukovec (approved):

PR #91, opened 2021-01-27
Fix headers.

nives113:


DarkoKukovec (commented):
Should debugging and testing chapters also be updated?

nives113:
> Should debugging and testing chapters also be updated?

Not sure, there are online markup viewers (probably a VS plugin too) and that is what I used to see if something is broken. Maybe some guidelines can be added (for example: rules not to use # H1, since the chapter already gets title from file name, filename conventions, etc).

DarkoKukovec:
Maybe, but in the debugging and testing cases, they have a h2 title that is equal to the chapter title, so we have it twice on the page.

fvoska:
If you need to debug this quickly, you can merge the changes to `staging` branch and see it deployed to https://beta.infinum.com/handbook/books/frontend/, so you can see the changes and experiment.

DarkoKukovec (approved):

PR #98, opened 2021-02-23
Nextjs same origin proxy

filipjakov:
As per @andreicek 's suggestions

### DIFF
--- a/React/Next/Development proxy.md
+++ b/React/Next/Development proxy.md
@@ -28,14 +28,37 @@ import { NextApiRequest, NextApiResponse } from 'next';
 
 let apiUrl: string;
 
-switch (process.env.PROXY_ENV) { // pick API endpoint depending on the PROXY_ENV, assign to apiUrl }
+switch (process.env.PROXY_ENV) { 
+  // pick API endpoint depending on the PROXY_ENV, assign to apiUrl 
+  case 'production':
+    apiUrl = 'https://production.example.com';
+    break;
+  case 'uat':
+    apiUrl = 'https://uat.example.com';
+    break;
+  default:
+    apiUrl = 'https://staging.example.com';
+}
 
 const proxy = createProxyMiddleware({
   target: apiUrl,
   changeOrigin: true,
   logLevel: 'debug',
   cookieDomainRewrite: 'localhost',
-  onProxyRes: (proxyRes: IncomingMessage) => { // You can manipulate the cookie here },
+  onProxyRes: (proxyRes: IncomingMessage) => { 
+    // You can manipulate the cookie here
+
+    if (!proxyRes.headers['set-cookie']) {
+      return;
+    }
+
+    // For example you can remove secure and SameSite security flags so browser can save the cookie in dev env
+    const adaptCookiesForLocalhost = proxyRes.headers['set-cookie'].map((cookie) =>
+      cookie.replace(/; secure/gi, '').replace(/; SameSite=None/gi, ''),
+    );
+
+    proxyRes.headers['set-cookie'] = adaptCookiesForLocalhost;
+  },
   onError: (err: Error) => console.error(err),
 });
 
--- a/React/Next/Development proxy.md
+++ b/React/Next/Development proxy.md
@@ -1,6 +1,6 @@
 ## Motivation
 
-When developing applications, sometimes you want to test your local changes against data from a different environment. For example: creating a production hotfix requires you to locally connect to the production API. Maybe even testing the solution on a different device e.g. a [phone connected to your machine](https://developers.google.com/web/tools/chrome-devtools/remote-debugging). That API could have http-only cookies that cannot be manipulated on the client and require a proxy that can intercept a request and do some modifications on it.
+When developing applications, you sometimes want to test your local changes against data from a different environment. For example, creating a production hotfix requires you to locally connect to the production API. Maybe even testing the solution on a different device, e.g. a [phone connected to your machine](https://developers.google.com/web/tools/chrome-devtools/remote-debugging). That API could have http-only cookies that cannot be manipulated on the client and require a proxy that can intercept some requests and do some modifications to them.
 
 ## The issue
 
--- a/React/Next/Development proxy.md
+++ b/React/Next/Development proxy.md
@@ -84,4 +84,4 @@ We can then run `PROXY_ENV=<insert-env> next dev` and then NextJS development in
 
 ## The implications
 
-The only problem with this approach is that the defined NextJS route (`pages/api/[[...slug]]`) can be reached in production since NextJS doesn't support conditional removal of API routes. We avoided the invocation of the proxy middleware by check if the current environment is NOT "development" and returned a 404 (can be changed to something more suitable).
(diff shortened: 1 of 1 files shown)


React/Next/Development proxy.md
```diff
@@ -0,0 +1,64 @@
...
+
+```js
+// pages/api/[[...slug]].ts
+import { IncomingMessage } from 'http';
+import { createProxyMiddleware } from 'http-proxy-middleware';
+import { NextApiRequest, NextApiResponse } from 'next';
+
+let apiUrl: string;
+
+switch (process.env.PROXY_ENV) { // pick API endpoint depending on the PROXY_ENV, assign to apiUrl }
```
DarkoKukovec, line 31: This should probably be a multiline comment to avoid commenting out the closing }

isBatak (approved):

fvoska:
You will also need to update frontend-handbook-private repo with this new file in the sidebar file structure.

React/Next/Development proxy.md
```diff
@@ -0,0 +1,87 @@
+## Motivation
+
+When developing applications, sometimes you want to test your local changes against data from a different environment. For example: creating a production hotfix requires you to locally connect to the production API. Maybe even testing the solution on a different device e.g. a [phone connected to your machine](https://developers.google.com/web/tools/chrome-devtools/remote-debugging). That API could have http-only cookies that cannot be manipulated on the client and require a proxy that can intercept a request and do some modifications on it.
```
fvoska, line 3:
```suggestion
When developing applications, you sometimes want to test your local changes against data from a different environment. For example, creating a production hotfix requires you to locally connect to the production API. Maybe even testing the solution on a different device, e.g. a [phone connected to your machine](https://developers.google.com/web/tools/chrome-devtools/remote-debugging). That API could have http-only cookies that cannot be manipulated on the client and require a proxy that can intercept some requests and do some modifications to them.
```

React/Next/Development proxy.md
```diff
@@ -0,0 +1,87 @@
...
+    externalResolver: true, // hide warning message
+  },
+};
+```
+
+We can then run `PROXY_ENV=<insert-env> next dev` and then NextJS development instance will point to the appropriate API endpoint.
+
+## The implications
+
+The only problem with this approach is that the defined NextJS route (`pages/api/[[...slug]]`) can be reached in production since NextJS doesn't support conditional removal of API routes. We avoided the invocation of the proxy middleware by check if the current environment is NOT "development" and returned a 404 (can be changed to something more suitable).
```
fvoska, line 87:
```suggestion
The only problem with this approach is that the defined NextJS route (`pages/api/[[...slug]]`) can be reached in production because NextJS doesn't support conditional removal of API routes. We can avoid the invocation of the proxy middleware by checking if the current environment is NOT "development" and return a 404 (can be changed to something more suitable).
```

fvoska (changes requested):

React/Next/Development proxy.md
fvoska, line 87: This part makes is sound like we are misusing this feature a bit. It seems like the original intended purpose of API Routes was not to have a local development proxy.

React/Next/Development proxy.md
```diff
@@ -0,0 +1,87 @@
...
+    apiUrl = 'https://uat.example.com';
+    break;
+  default:
+    apiUrl = 'https://staging.example.com';
+}
+
+const proxy = createProxyMiddleware({
+  target: apiUrl,
+  changeOrigin: true,
+  logLevel: 'debug',
```
andreicek, line 46: this should be controller over a global log level which you should have in your app. debug log can be huge and a privacy issue on production sites.

React/Next/Development proxy.md
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    // For example you can remove secure and SameSite security flags so browser can save the cookie in dev env
+    const adaptCookiesForLocalhost = proxyRes.headers['set-cookie'].map((cookie) =>
+      cookie.replace(/; secure/gi, '').replace(/; SameSite=None/gi, ''),
+    );
+
+    proxyRes.headers['set-cookie'] = adaptCookiesForLocalhost;
+  },
+  onError: (err: Error) => console.error(err),
```
andreicek, line 62: you should detail how (if) to recover from an error, even if console log is enough for an example.

React/Next/Development proxy.md
```diff
@@ -0,0 +1,87 @@
...
+});
+
+export default function handler(req: NextApiRequest, res: NextApiResponse<unknown>) {
+  // Don't allow requests to hit the proxy when not in development mode
```
andreicek, line 66: can you skip inclusion of this file if you're not in a production build? webpack has a feature for this - don't know if iti's supported in next.js

React/Next/Development proxy.md
```diff
@@ -0,0 +1,87 @@
...
+
+```js
+// pages/api/[[...slug]].ts
+import { IncomingMessage } from 'http';
+import { createProxyMiddleware } from 'http-proxy-middleware';
+import { NextApiRequest, NextApiResponse } from 'next';
+
+let apiUrl: string;
+
+switch (process.env.PROXY_ENV) {
```
andreicek, line 31: it would be good to default to NODE_ENV if PROXY_ENV is not set

andreicek (changes requested):
thank you for the effort

React/Next/Development proxy.md
filipjakov: @fvoska that is kind of right, but there is no other way atm if you want the proxy on the same server. [Redirects](https://nextjs.org/docs/api-reference/next.config.js/redirects) looked promising, but you can use them only for redirection, and can't manipulate request/response :( when they add that, this whole API routes thing becomes obsolete

React/Next/Development proxy.md
filipjakov: Great point!

What do you think about the following:

```js
...
const proxy = createProxyMiddleware({
  target: apiUrl,
  changeOrigin: true,
  logLevel: process.env.NODE_ENV === 'development' ? 'debug' : 'silent',
 ...
```

React/Next/Development proxy.md
filipjakov: @andreicek googled but didn't find anything useful. NextJS uses some kind of "proprietary" webpack (correct me if I am wrong @isBatak ), so not all webpack possibilities are doable with NextJS.

React/Next/Development proxy.md
filipjakov: TIL you can have logical conditions in a switch condition :D

what do you think about this @andreicek?

```js
switch (process.env.PROXY_ENV || process.env.NODE_ENV) {...} 
```

Also, the default value in the switch should cover the default API (staging), so I'm not sure if it's necessary.

React/Next/Development proxy.md
andreicek: ok

andreicek (approved):
good to merge on my side

DarkoKukovec (approved):

fvoska (approved):

PR #103, opened 2021-02-24
Feature - accessibility handbook

kristian240:


Accessibility.md
```diff
@@ -0,0 +1,32 @@
...
+Why we should we focus on a11y?
+What are the benefits?
+Maybe some cons?
+
+## Screen reader
+
+Mac's Voice Over (and alternatives)
+Window's screen readers
+
+## Semantics
```
kristijanbambir, line 14: I would perhaps rename it to "Semantics in HTML".

Accessibility.md
```diff
@@ -0,0 +1,32 @@
...
+
+## ARIA
+
+What is ARIA?
+How to use it?
+Links to good examples of using ARIA
+
+## Examples
+
+Do we need this section?
```
kristijanbambir, line 28: Here I would expect example implementations of common components we develop in applications (buttons, accordians, carousels, etc.) that showcase how to make them more accessible.

kristijanbambir (approved):
The draft looks good to me 👍

kristian240:
The day has come 🎉 [Here is v0 of an accessibility handbook.](https://docs.google.com/document/d/1DRZ9S56SqdNXLVIibmc45H1WgU7B1i9meRRQHfhWBsA/edit?usp=sharing)

With more knowledge on accessibility, I changed few sections that were planned initially in this draft but I think all planned questions are covered.

In agreement with @kristijanbambir, I created this in Google Docs so it is easier to comment and edit the code. Once I get the green light, I'll convert this document to a markdown.

Accessibility.md
```diff
@@ -0,0 +1,32 @@
...
+
+Mac's Voice Over (and alternatives)
+Window's screen readers
+
+## Semantics
+
+What is semantics? (short paragrphs)
+List of semantic elements (or link to extrnal list)
+How to structure a page?
+
```
isBatak, line 19: What about Accessible Colors?

Storybook also has a plugin for a11y https://storybook.js.org/addons/@storybook/addon-a11y maybe its cool thing to mention.

kristian240: > What about Accessible Colors?

I am not really sure if this is something we should really worry about. I mean, yeah, if we notice that the contrast is low (white bgColor and light gray fontColor) we should mention this, but often we just implement what we get from designers.

Maybe we can add the A11y section to the pre-production checklist?

kristian240: > Storybook also has a plugin for a11y https://storybook.js.org/addons/@storybook/addon-a11y maybe its cool thing to mention.

I agree, this looks cool! Maybe we add a new section with extensions, addons, plugins for storybook, browsers, IDEs? What do you think?

kristijanbambir (approved):
As discussed in person, this is good and well written as it explains the reasons and motivation behind a11y. What is missing is a section/chapter with hands on things to think about and use when creating a web application. The semantic HTML section should be expanded to clearly state to:
1. Use button elements for actions that trigger some logic, and use anchor elements for navigation actions.
2. Set a button role in case a click event is handled on an element that is not a button.
3. Use section elements for page sections.
4. …
As agreed, please add this as a separate topic.

kristijanbambir (approved):
As discussed in person, this chapter is well written and provides motivation and reasons behind a11y. As agreed, please add another chapter with hands-on examples of what every application must follow, and some more advanced things to follow.

PR #107, opened 2021-03-31
Add react hooks chapter

adrianbicanic:
This is a draft version of the React Hooks chapter.

**Notes:**
- it wasn't clear which branch should be used, so I created a separate file (for now) and the branch was created from `master` 
- There where some late additions, and restructurings of the original version I was working on, so there might be typos I didn't catch; also, some additional examples can be added, and some are a bit raw (suggestions are appreciated)

There are things we should probably include as specific subchapters that we don't have here yet (`useReducer`, `useRef`, Testing hooks, etc.)

@isBatak recently sent some links to check out; I didn't have much time for before the initial PR. There could be stuff that can be added or that are expanding/changing existing stuff.


React/React hooks.md
```diff
@@ -0,0 +1,386 @@
...
+
+  return <Button onClick={onButtonClick} />;
+}
+```
+
+#### Before you use memoization
+
+Have in mind that React is really good at optimizing rerenders by default.
+
+You might get tempted to wrap values and functions with `useMemo` and `useCallback` all the time, but in many of these cases, you don't really need it, and you might even make your app performance and file size worse. These calculation's can be expensive and you could end up using more memory than you would without them and make you code more complicated to read and mantain.
```
DarkoKukovec, line 384:
```suggestion
You might get tempted to wrap values and functions with `useMemo` and `useCallback` all the time, but in many of these cases, you don't really need it, and you might even make your app performance and file size worse. These calculations can be expensive and you could end up using more memory than you would without them and make you code more complicated to read and maintain.
```

React/React hooks.md
```diff
@@ -0,0 +1,386 @@
...
+    ...
+  }, [...])
+
+  return <LargeList onListItemClick={onListItemClick} />;
+}
+```
+
+#### You don't have to avoid "inlining" non-expensive functions
+
+React is good at optimizing, so if you prematurly decide to wrap a function inside a `useCallback` and declare it in the component scope, you might get slower performance in many cases, because how memoization can be expensive. In most cases, even if you improve performance, it will be insignificant in comparison to making your code less clean and harder to mantain.
```
DarkoKukovec, line 358: The first sentence seems weird and too long. Maybe split it into two and start the second one with "Because memoization can be expensive"

React/React hooks.md
```diff
@@ -0,0 +1,386 @@
...
+
+  // ...
+}
+```
+
+Assuming that `propA` and `propB` are primitive (or memoized, non-primitive) values, with `useMemo` you are ensuring that non-primitive references are retained throughout rerenders, until `propA` or `propB` have changed.
+
+```jsx
+export const MyComponent: FC = ({ propA, propB }) => {
+  const memoizedObject = useMemo(() => {
```
isBatak, line 243: This is not a good example.
This is the correct way of doing it.
```
useEffect(() => {
   const object = {
     a: propA,
     b: propB,
   };
   // do something wth the object
}, [propA, propB])
```

React/React hooks.md
```diff
@@ -0,0 +1,386 @@
...
+
+  if (!id.current) {
+    id.current = generateId();
+  }
+
+  // ...
+}
+```
+
+If you want the component to rerender when the value is set:
```
isBatak, line 310: You should never want that.
This will trigger two renders.
Remember the React hooks flow.
https://github.com/donavon/hook-flow
`useEffect` is executed at the end, when the first paint is already done, so this will trigger the second render, which is bad for performance.
This is a BAD example and we should note this in the handbook.
Don't ever do this!

isBatak: That reminds me that we should add a chapter about hooks flow and include this chart

React/React hooks.md
```diff
@@ -0,0 +1,386 @@
...
+}
+```
+
+
+```jsx
+// GOOD
+export const MyComponent: FC = () => {
+  const [count, setCount] = useState(0);
+
+  const increment = useCallBack((stepSize = 1) => { setState((previousCount) => previousCount + stepSize) });
```
isBatak, line 374: You are missing deps array

isBatak:
```
useCallBack
       ^
       └── lowercase `b`
```

React/React hooks.md
```diff
@@ -0,0 +1,386 @@
...
+}
+```
+
+#### Before you use memoization
+
+Have in mind that React is really good at optimizing rerenders by default.
+
+You might get tempted to wrap values and functions with `useMemo` and `useCallback` all the time, but in many of these cases, you don't really need it, and you might even make your app performance and file size worse. These calculation's can be expensive and you could end up using more memory than you would without them and make you code more complicated to read and mantain.
+
+If it's not obvious that memoization is necessary, profile your app performance without it first, using React dev tools, and then optimize if necessary.
```
isBatak, line 386: Maybe we could add some screenshots of the profiler

isBatak (changes requested):

React/React hooks.md
isBatak: Fixed in #109

React/React hooks.md
isBatak: @DarkoKukovec I think it's better now

React/React hooks.md
isBatak: This is fixed in #109

React/React hooks.md
```diff
@@ -0,0 +1,754 @@
+### React hooks
```
DarkoKukovec, line 1: You should fix the headings. The title will be shown automatically based on the file name. All other headings should start from h2

React/React hooks.md
```diff
@@ -0,0 +1,754 @@
...
+}
+```
+
+You can also store the value in a ref:
+> Note: updating ref values will not trigger a re-render.
+
+```jsx
+type Result<T> = { v: T };
+
+function useConstant<T>(fn: () => T): T {
```
isBatak, line 444: I'm thinking to remove `useConstant` example.
 The explanation is here in the thread https://twitter.com/sstur_/status/1384934568285900806
 Maybe I could leave it and write an explanation of what is wrong...
Consider this:
1. render starts
2. you update the ref
3. react concurrent mode has to throw away the work and doesn't commit
You now have a ref that isn't the latest value, it's a value for a render that was never committed.

DarkoKukovec (approved):

isBatak (approved):

PR #109, opened 2021-04-09
Add react hooks chapter - extended

isBatak:


DarkoKukovec (approved):

React/React hooks.md
```diff
@@ -214,99 +343,205 @@ export const MyComponent: FC = ({ propA, propB }) => {
...
+}
+
+export const Component: FC = () => {
+  const isClient = useIsClient();
+
+  return <div>{isClient ? 'Client' : 'Server'}</div>
+}
+```
+
+#### Wrap functions with `useCallback` it they can cause large subtrees to re-render too often
```
kristian240, line 544:
```suggestion
#### Wrap functions with `useCallback` if they can cause large subtrees to re-render too often
```

React/React hooks.md
```diff
@@ -341,29 +576,180 @@ export const MyList: FC = () => {
...
+```
 
-  const increment = useCallback((stepSize = 1) => { setState((previousCount) => previousCount + stepSize) });
+You can see that our component becomes messy and hard to understand. There is a lot of code in the body of the function which increase our [cognitive load](https://en.wikipedia.org/wiki/Cognitive_load). To fix this issue we can **encapsulate** our **elements of concern** into a separate _chunks of work_, i.e. custom hooks.
 
-  return <Button onClick={onButtonClick} />;
+The Solution:
+
+```jsx
+const useCount = (initialState = 0) => {
```
kristian240, line 667: Missing stepSize in function params

React/React hooks.md
```diff
@@ -341,29 +576,180 @@ export const MyList: FC = () => {
...
+}
+
+const useInput = (initialState = '') => {
+  const [state, setState] = useState(() => initialState);
+
+  const handlers = useMemo(() => ({
+    handleInputChange: (event) => {
+      setState(event.target.value);
+    }
+  })), []);
```
kristian240, line 695: Are we missing initialState in dep array?

```suggestion
  })), [initialState]);
```

kristian240 (commented):
👏🏼

React/React hooks.md
```diff
@@ -341,29 +576,180 @@ export const MyList: FC = () => {
...
 
+#### Hooks encapsulation
+
+Common problem with hooks is that the components using them could go out of control and become unreadable and messy. 
+That happens when you have multiple `useEffect` and `useCallback` in you Function Component body, and that happens often if you are building real world product. To overcome this problem we can do the same thing as we would do the Class Component, split things in smaller chunks of logic and put them somewhere else, i.e. private methods of Class Component or custom hooks in Function Components.
 
+The Problem:
+
+This is the continuation of previous section, but we will expand it with some additional requirements.  
+We need to add input field in which we can input the thing we are counting, and add decrement and reset handlers for counter.
```
adrianbicanic, line 610: I'm guessing this is what you meant here.

```suggestion
We need to add an input field for setting the description of the value that we are counting, with decrement and reset handlers for the counter.
```

React/React hooks.md
```diff
@@ -1,7 +1,68 @@
 ### React hooks
 
-#### Using hook dependency arrays
-Dependency arrays are a way to trigger re-firing of hooks only when it's dependencies change. React will keep the values generated by a hook stored in memory and then run it when a change of a dependency has occured. This applies to all hooks that can have dependencies.
+React Hooks were introduced in version 16.8.0 as Function Component counterpart of Class Component Lifecycle.
```
adrianbicanic, line 3: Correct me if you think we need capitalisation here.

```suggestion
React Hooks were introduced in version 16.8.0 as function component counterpart of class component lifecycles.  
```

React/React hooks.md
```diff
@@ -1,7 +1,68 @@
 ### React hooks
 
-#### Using hook dependency arrays
-Dependency arrays are a way to trigger re-firing of hooks only when it's dependencies change. React will keep the values generated by a hook stored in memory and then run it when a change of a dependency has occured. This applies to all hooks that can have dependencies.
+React Hooks were introduced in version 16.8.0 as Function Component counterpart of Class Component Lifecycle.  
+For more information, see the official [React Hooks API Reference](https://reactjs.org/docs/hooks-reference.html).
+
+#### Hooks flow
+Before we start using hooks in React we need to understand the flow.
```
adrianbicanic, line 7:
```suggestion
Before starting, it is important to understand the flow of react hooks.
```

These suggestions are also "meh", but we should  somehow avoid repeating ourselves too much in consecutive sentences.

React/React hooks.md
```diff
@@ -1,7 +1,68 @@
...
+
+Here is a flow diagram that explains the new flow of a Hooks component.
```
adrianbicanic, line 9:
```suggestion
Here is a diagram that explains it visually:


![hooks-flow](https://raw.githubusercontent.com/donavon/hook-flow/master/hook-flow.png)
```

React/React hooks.md
```diff
@@ -1,7 +1,68 @@
...
+<br/>
+<br/>
+![hooks-flow](https://raw.githubusercontent.com/donavon/hook-flow/master/hook-flow.png)
+
+> Most important thing here is to notice how "Run Effects" phase is executed last!
+
+Here is the code example that explains this flow.
```
adrianbicanic, line 16:
```suggestion
This is a code example that explains this flow:
```

React/React hooks.md
```diff
@@ -1,7 +1,68 @@
...
+  );
+}
+```
+
+In the "Mount" phase you will see:
+```
+Prev: undefined
+Current: 0
+```
+This is because `useEffect` is called after the first render and also assigning values to `ref` does not trigger re-render, instead, the value is populated and waiting for next update phase.
```
adrianbicanic, line 48:
```suggestion
This is because `useEffect` is called after the first render and because assigning a value to `ref` does not trigger a re-render - instead, the value is populated and waiting for the next update phase.
```

React/React hooks.md
```diff
@@ -1,7 +1,68 @@
...
+
+When button is clicked React will trigger `Update` phase and result will be:
```
adrianbicanic, line 50:
```suggestion
When `button` is clicked, React will trigger the `update` phase and the result will be:
```

React/React hooks.md
```diff
@@ -13,19 +74,68 @@ export const MyComponent: FC = ({ numberProp, stringProps }) => {
 }
 ```
 
-For non-primitive values like objects, arrays and functions, React will do a reference comparison, instead of comparing them with their previous values.
+For non-primitive values like objects, arrays and functions, React will do a reference comparison using [Object.is()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is). This means that we must provide the same reference of `objectProp`, `arrayProp`, `functionProp` to `Child` on every re-render if we don't want to trigger `useEffect` every time. This can be achieved by extracting them outside component function or wrapping them in `useMemo` or `useCallback`.
```
adrianbicanic, line 77:
```suggestion
For non-primitive values like objects, arrays and functions, React will do a reference comparison using [Object.is()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is). This means that we must provide the same reference of `objectProp`, `arrayProp`, `functionProp` to `Child` on every re-render if we don't want to trigger `useEffect` every time. This can be achieved by extracting them outside of the component or wrapping them in `useMemo` or `useCallback`.  
```

React/React hooks.md
```diff
@@ -13,19 +74,68 @@ export const MyComponent: FC = ({ numberProp, stringProps }) => {
...
+    <Child 
+      objectProp={objectProp} 
+      arrayProp={arrayProp} 
+      functionProp={functionProp} 
+    />
+    // some other components...
+  );
+}
+
+// FUTURE with Records & Tuples (finger crossed)
```
adrianbicanic, line 124:
```suggestion
// FUTURE with Records & Tuples
```

or

```suggestion
// FUTURE with Records & Tuples (fingers crossed)
```

adrianbicanic:
@isBatak I like the parts that you added. Most of my comments are related to syntax and grammar. I might have missed some things. 

We should definitely check all of it when we are done, so we see if we were consistent. There's a lot of stuff here now, so it could be easy to miss.

React/React hooks.md
isBatak: noup! read it again :D

React/React hooks.md
```diff
@@ -341,29 +576,180 @@ export const MyList: FC = () => {
...
+      <StepButton onClick={increment} stepSize={2} />
+    </div>
+  );
 }
 ```
 
+#### Hooks encapsulation
+
+Common problem with hooks is that the components using them could go out of control and become unreadable and messy. 
+That happens when you have multiple `useEffect` and `useCallback` in you Function Component body, and that happens often if you are building real world product. To overcome this problem we can do the same thing as we would do the Class Component, split things in smaller chunks of logic and put them somewhere else, i.e. private methods of Class Component or custom hooks in Function Components.
```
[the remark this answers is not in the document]
isBatak: or `multiple invocations of`

React/React hooks.md
kristian240: Ahaaa 🙈

PR #111, opened 2021-04-16
React hooks - Add more chapters

isBatak:
WIP


React/React hooks.md
```diff
@@ -732,6 +732,291 @@ Useful links:
...
+    <div>
+      <BreadCrumbs />
+    </div>
+  );
+}
+```
+
+The Problem ⚡  
+
+There are two use cases, the "data-fetching" and "displaying breadcrumbs". Both are updated with an useEffect hook. This single useEffect hooks will run when the fetchData and updateBreadcrumbs functions or the location changes. The main problem is now, we also call the fetchData function when the location changes. This might be a side effect we haven't thought of.
```
DarkoKukovec, line 842:
```suggestion
There are two use cases, the "data-fetching" and "displaying breadcrumbs". Both are updated with an useEffect hook. This single useEffect hooks will run when the fetchData and updateBreadcrumbs functions on the location changes. The main problem is now, we also call the fetchData function when the location changes. This might be a side effect we haven't thought of.
```

DarkoKukovec: tbh, I don't exactly understand the point you're trying to make here. It should probably be more clear that fetchData doesn't depend on the path.

isBatak: I'm still working on this one. Should have added WIP or draft.
Chapters:
#### Handling actions via useEffect 
and
#### Single responsibility useEffects
are copied from some article, I'll do my own version of examples and try to better explain the problem.
This was just the layout of the idea.

isBatak:
I'm closing this PR because we decided to use SWR for data fetching, so all of these edge-cases becomes irrelevant.

PR #113, opened 2021-04-19
Add how to export SVG section to handbook

kristian240:
Added section where I explain how to export SVG from Figma/Zeplin.

### DIFF
--- a/Assets.md
+++ b/Assets.md
@@ -1,6 +1,6 @@
-# SVGs
+## SVGs
 
-## 1. Export
+### 1. Export
 
 When we export an icon from the Figma or Zeplin, we should export the icon container instead of the actual icon. _Check the image_
 ![image](https://user-images.githubusercontent.com/55184443/115124509-edfa2e80-9fc2-11eb-8386-fa29ad4c0548.png)
@@ -8,15 +8,15 @@ In this case, we will export `Menu` as SVG instead of the `Icon` element.
 
 > With this, every icon we export will have the exact box dimensions, which is easier to maintain in the long run.
 
-## 2. Optimization
+### 2. Optimization
 
 Before adding a new icon to our codebase, we should optimize it using [SVGOMG](https://jakearchibald.github.io/svgomg/) tool. [Here are settings](https://gist.github.com/kristian240/bf7be2570e7cc8074718484130d5ae2e) I use for a while now, and at the time, when I was configuring my settings, this would give me the best output. Feel free to reuse them.
 
-## 3. Add to codebase
+### 3. Add to codebase
 
 After we exported and optimized the icon, we can add it to our codebase. We will add this new icon in one of two directories:
 
-### a. `src/assets/icons`
+#### a. `src/assets/icons`
 
 We should add an icon in this folder if the icon uses only one color and will be used as an icon in buttons/dropdowns/... such as `StarIcon`, `ChevronIcon`, etc. One last change needed before we add it to the codebase is to change the `fill` color to `currentColor`. This way, a developer can use any color with a simple `color: ${someColor}` style in CSS.
 
@@ -34,7 +34,7 @@ const AddButton = () => (
 
 > Make sure that SVG element has only the `viewBox` property (without `width` and `height`). Context, where the icon is used, should provide `width` and `height` to the icon.
 
-### b. `public/images`
+#### b. `public/images`
 
 If the image is using complex colors gradients/multiple colors/... it probably contains in this category.
 
--- a/Assets.md
+++ b/Assets.md
@@ -18,7 +18,12 @@ After we exported and optimized the icon, we can add it to our codebase. We will
 
 #### a. `src/assets/icons`
 
-We should add an icon to this folder if it uses only one color and will be used in components like buttons and dropdowns. The `fill` property of the icon source should be changed to `currentColor`. This way its color can be changed by setting the `color: ${someColor}` style in CSS.
+Before using this a SVG icon this way, few things are mandatory:
+
+- Make sure that the SVG element has only the `viewBox` property (without `width` and `height`). The context surrounding the icon should define its dimensions.
+- Sometimes, designers use `<mask>` element in SVGs. Those elemens are referenced by ID and if you use multiple SVGs per HTML document, mulitple `mask` elements will have same ID and as a result, SVGs won't work as expected. Often these `mask` elements are simple to resolve by hand so make sure to remove them.
+
+We should add an icon to this folder if it uses only one color and will be used in components like buttons and dropdowns and/or will be inlined with the code. The `fill` property of the icon source should be changed to `currentColor`. This way its color can be changed by setting the `color: ${someColor}` style in CSS.
 
 Example in code:
 
@@ -31,8 +36,6 @@ const AddButton = () => (
   </Button>
 );
 ```
-
-> Make sure that the SVG element has only the `viewBox` property (without `width` and `height`). The context surrounding the icon should define its dimensions.
 
 #### b. `public/images`
 
--- a/Assets.md
+++ b/Assets.md
@@ -21,7 +21,7 @@ After we exported and optimized the icon, we can add it to our codebase. We will
 Prior to using an SVG icon this way, check the following:
 
 - Make sure that the SVG element has only the `viewBox` property (without `width` and `height`). The context surrounding the icon should define its dimensions.
-- Sometimes, designers use the `<mask>` element in SVGs. Those elements are referenced by an ID, and if you use multiple SVGs per an HTML document, multiple `mask` elements will have the same ID. As a result, SVGs won't work as expected. Often these `mask` elements are simple to resolve by hand so make sure to remove them.
+- Sometimes, designers use the `<mask>` element in SVGs. Those elements are referenced by an ID, and if you use multiple SVGs per an HTML document, multiple `mask` elements will have the same ID. As a result, SVGs won't work as expected. Often these `mask` elements are simple to resolve manually so after resolving them make sure to remove them.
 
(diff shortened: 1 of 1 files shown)


DarkoKukovec (approved):
You should remove the `# SVGs` title because the title will be automatically added (based on the filename) when deployed.

Assets.md
```diff
@@ -0,0 +1,52 @@
...
+    <Icon as={AddIcon} color='primary' />
+  </Button>
+);
+```
+
+> Make sure that SVG element has only the `viewBox` property (without `width` and `height`). Context, where the icon is used, should provide `width` and `height` to the icon.
+
+#### b. `public/images`
+
+If the image is using complex colors gradients/multiple colors/... it probably contains in this category.
```
isBatak, line 39:
```suggestion
If the image is using complex colors gradients/multiple colors/... it probably belongs to this category.
```
or
```suggestion
If the image is using complex colors gradients/multiple colors/... it probably falls into this category.
```

Assets.md
```diff
@@ -0,0 +1,52 @@
+## SVGs
+
+### 1. Export
+
+When we export an icon from the Figma or Zeplin, we should export the icon container instead of the actual icon. _Check the image_
```
kristijanbambir, line 5:
```suggestion
When we export an icon from Figma or Zeplin, we should export the icon container instead of the actual icon. _Check the image_
```

Assets.md
```diff
@@ -0,0 +1,52 @@
...
+![image](https://user-images.githubusercontent.com/55184443/115124509-edfa2e80-9fc2-11eb-8386-fa29ad4c0548.png)
+In this case, we will export `Menu` as SVG instead of the `Icon` element.
+
+> With this, every icon we export will have the exact box dimensions, which is easier to maintain in the long run.
+
+### 2. Optimization
+
+Before adding a new icon to our codebase, we should optimize it using [SVGOMG](https://jakearchibald.github.io/svgomg/) tool. [Here are settings](https://gist.github.com/kristian240/bf7be2570e7cc8074718484130d5ae2e) I use for a while now, and at the time, when I was configuring my settings, this would give me the best output. Feel free to reuse them.
```
kristijanbambir, line 13:
```suggestion
Before adding a new icon to our codebase, we should optimize it. [`svgo`](https://github.com/svg/svgo) is a go-to tool for this purpose, with [SVGOMG](https://jakearchibald.github.io/svgomg/) as its hosted UI. [Here are the preferred settings](https://gist.github.com/kristian240/bf7be2570e7cc8074718484130d5ae2e) that should give the optimal output.
```

Assets.md
```diff
@@ -0,0 +1,52 @@
...
+
+### 3. Add to codebase
+
+After we exported and optimized the icon, we can add it to our codebase. We will add this new icon in one of two directories:
+
+#### a. `src/assets/icons`
+
+We should add an icon in this folder if the icon uses only one color and will be used as an icon in buttons/dropdowns/... such as `StarIcon`, `ChevronIcon`, etc. One last change needed before we add it to the codebase is to change the `fill` color to `currentColor`. This way, a developer can use any color with a simple `color: ${someColor}` style in CSS.
```
kristijanbambir, line 21:
```suggestion
We should add an icon to this folder if it uses only one color and will be used in components like buttons and dropdowns. The `fill` property of the icon source should be changed to `currentColor`. This way its color can be changed by setting the `color: ${someColor}` style in CSS.
```

Assets.md
```diff
@@ -0,0 +1,52 @@
...
+import AddIcon from 'src/assets/icons/icon-add.svg';
+
+const AddButton = () => (
+  <Button>
+    <Icon as={AddIcon} color='primary' />
+  </Button>
+);
+```
+
+> Make sure that SVG element has only the `viewBox` property (without `width` and `height`). Context, where the icon is used, should provide `width` and `height` to the icon.
```
kristijanbambir, line 35:
```suggestion
> Make sure that the SVG element has only the `viewBox` property (without `width` and `height`). The context surrounding the icon should define its dimensions.
```

Assets.md
```diff
@@ -0,0 +1,52 @@
...
+
+#### b. `public/images`
+
+If the image is using complex colors gradients/multiple colors/... it probably contains in this category.
+
+> Before adding it, you can check if this image could be smaller in size using `png` or `jpg` image format. If this is the case, use that instead of `SVG`.
```
kristijanbambir, line 41: Actually, this depends on the use-case. If the image container is variable in size, using an SVG you would not need to worry about providing a large enough PNG for example.

kristijanbambir (changes requested):

Assets.md
kristian240: I'll add a note for this

fvoska:
I think it would be good to add a note somewhere that if you plan to inline the SVGs, it is important that there are no IDs or classes in SVGs as that can cause issues if more than one SVG is inlined and if they have some common IDs and classes.

Kinda related - often when we export SVG we get unnecessary IDs that refer to some `<mask>` elements. SVGO doesn't remove those but for simple icons it is usually not much work to remove masks and IDs.

Assets.md
```diff
@@ -0,0 +1,52 @@
...
+);
+```
+
+> Make sure that the SVG element has only the `viewBox` property (without `width` and `height`). The context surrounding the icon should define its dimensions.
+
+#### b. `public/images`
+
+If the image is using complex colors gradients/multiple colors/... it probably belongs to this category.
+
+> Before adding it, you can check if this image could be smaller in size using `png` or `jpg` image format. If this is the case, use that one instead of `SVG`. If you opt in for this, be careful if this icon will be used in large formats, `png` and `jpg` could strech and result as a pixelated image.
```
kristijanbambir, line 41:
```suggestion
> Before adding it, you can check if this image could be smaller in size using PNG or JPG image formats. If this is the case, and the image doesn't need to scale dynamically, use that one instead of the SVG. If instead the image needs to change dimensions based on the viewport width for instance, use the SVG.
```

kristijanbambir (changes requested):

isBatak:
We should also add Angular examples.

Assets.md
```diff
@@ -0,0 +1,52 @@
...
+After we exported and optimized the icon, we can add it to our codebase. We will add this new icon in one of two directories:
+
+#### a. `src/assets/icons`
+
+We should add an icon to this folder if it uses only one color and will be used in components like buttons and dropdowns. The `fill` property of the icon source should be changed to `currentColor`. This way its color can be changed by setting the `color: ${someColor}` style in CSS.
+
+Example in code:
+
+```jsx
+import AddIcon from 'src/assets/icons/icon-add.svg';
```
isBatak, line 26: We should add a note somewhere that this line will only work if we configure `next-react-svg` or `svg-react-loader` for the storybook.

Also, what's with the Angular env?

andreicek:
@kristian240 what the status of this? this PR is getting mentioned in other chapters so i'm eager to get this merged asap.

Assets.md
kristian240: I added a not.

Regarding angular: https://infinum.slack.com/archives/CDYJW0327/p1624439103000800

kristian240:
@andreicek this is ready for the final check. Only missing is more examples (Angular, Vue?

Assets.md
```diff
@@ -0,0 +1,59 @@
...
+
+When we export an icon from Figma or Zeplin, we should export the icon container instead of the actual icon. _Check the image_
+![image](https://user-images.githubusercontent.com/55184443/115124509-edfa2e80-9fc2-11eb-8386-fa29ad4c0548.png)
+In this case, we will export `Menu` as SVG instead of the `Icon` element.
+
+> With this, every icon we export will have the exact box dimensions, which is easier to maintain in the long run.
+
+### 2. Optimization
+
+Before adding a new icon to our codebase, we should optimize it. [`svgo`](https://github.com/svg/svgo) is a go-to tool for this purpose, with [SVGOMG](https://jakearchibald.github.io/svgomg/) as its hosted UI. [Here are the preferred settings](https://gist.github.com/kristian240/bf7be2570e7cc8074718484130d5ae2e) that should give the optimal output.
```
kristijanbambir, line 13:
```suggestion
Before adding a new icon to our codebase, we should optimize it. [`svgo`](https://github.com/svg/svgo) is the go-to tool for this purpose, with [SVGOMG](https://jakearchibald.github.io/svgomg/) as its hosted UI. [Here are the preferred settings](https://gist.github.com/kristian240/bf7be2570e7cc8074718484130d5ae2e) that should give the optimal output.
```

Assets.md
```diff
@@ -0,0 +1,59 @@
...
+
+### 3. Add to codebase
+
+After we exported and optimized the icon, we can add it to our codebase. We will add this new icon in one of two directories:
+
+#### a. `src/assets/icons`
+
+Before using this a SVG icon this way, few things are mandatory:
```
kristijanbambir, line 21:
```suggestion
Prior to using an SVG icon this way, check the following:
```

Assets.md
```diff
@@ -0,0 +1,59 @@
...
+
+- Make sure that the SVG element has only the `viewBox` property (without `width` and `height`). The context surrounding the icon should define its dimensions.
+- Sometimes, designers use `<mask>` element in SVGs. Those elemens are referenced by ID and if you use multiple SVGs per HTML document, mulitple `mask` elements will have same ID and as a result, SVGs won't work as expected. Often these `mask` elements are simple to resolve by hand so make sure to remove them.
```
kristijanbambir, line 24:
```suggestion
- Sometimes, designers use the `<mask>` element in SVGs. Those elements are referenced by an ID, and if you use multiple SVGs per an HTML document, multiple `mask` elements will have the same ID. As a result, SVGs won't work as expected. Often these `mask` elements are simple to resolve by hand so make sure to remove them.
```
-----
>Often these `mask` elements are simple to resolve by hand so make sure to remove them.

This is a bit misleading. Should they be removed or resolved in some other way?

kristijanbambir (changes requested):

kristijanbambir (approved):

PR #115, opened 2021-05-07
Project structure refinement

isBatak:


React/Project structure.md
```diff
@@ -34,31 +34,202 @@ src
...
+    <th>Parts</th>
+    <th>Description</th>
+  </tr>
+  <tr>
+    <td>Card</td>
+    <td>
+      `Card`,  `CardImage`,  `CardImageOverlay`,  `CardTitle`, `CardDescription`, ...
+    </td>
+    <td>
+      From these parts, you'll be able to compose multiple more specific __molecules__ like `ProductCard` or `UserCard`.
```
DarkoKukovec, line 101:
```suggestion
      From these parts, you'll be able to compose multiple more specific **_molecules_** like `ProductCard` or `UserCard`.
```

DarkoKukovec (approved):

React/Project structure.md
```diff
@@ -79,6 +266,8 @@ Example:
     └── WelcomeCard.elements.ts
 ```
 
+> Moving things to`.elements.tsx` should be the last step in the development process and it should only be used for organizational purposes, i.e. when the main component becomes cluttered and unreadable.
```
mamihelj09, line 269: Not sure about this, I think that every components should have`.elements.tsx` instead of having styles in main file. Almost every component will have some kind of wrapper or something.

andreicek (approved):
👏

PR #121, opened 2021-05-20
How to build core component

isBatak:
This chapter explains how to create Compound component.


React/Recipes/How to build core component.md
```diff
@@ -0,0 +1,639 @@
...
+);
+```
+
+### When this is OK?
+
+When we are building a small one-off components which doesn't contain a lot of code and  are not shared across components.  
+
+### Why this approach can become a problem?
+
+This example is not so bad as it seems at firs because we don't have to name components (naming is hard), but as this component grows and new features are added, inline style will become a problem.
```
gbareza, line 101:
```suggestion
This example is not as bad as it seems at first because we don't have to name components (naming is hard), but as this component grows and new features are added, the inline style will become a problem.
```

React/Recipes/How to build core component.md
```diff
@@ -0,0 +1,639 @@
...
+
+```tsx
+const styles = useMultiStyleConfig("Card", props);
+```
+
+> [API Docs](https://chakra-ui.com/docs/styled-system/theming/component-style#usemultistyleconfig-api)
+
+#### Using StylesProvider
+
+In the rood component we can access the whole style config and provide it to all the parts via `StylesProvider` context provider. This way we can for example set the variant only on the root element and all the parts will inherit that variant styles.
```
gbareza, line 515:
```suggestion
In the root component, we can access the whole style config and provide it to all the parts via `StylesProvider` context provider. This way we can, for example, set the variant only on the root element and all the parts will inherit that variant styles.
```

React/Recipes/How to build core component.md
```diff
@@ -0,0 +1,639 @@
...
+});
+
+if (__DEV__) {
+  Card.displayName = "Card";
+}
+
+export const CardImage = forwardRef<ImageProps, "img">((props, ref) => {
+  const { image } = useStyles();
+
+  return <Image ref={ref} __css={image} {...props} />;
```
isBatak, line 599:
```suggestion
  return <Image ref={ref} {...image} {...props} />;
```

kristian240 (approved):

PR #127, opened 2021-06-21
Add section for icon implementation

SusanjRaul:
This PR depends on the PR of @kristian240 (#113).

### DIFF
--- a/React Native/Icons.md
+++ b/React Native/Icons.md
@@ -1,0 +1,45 @@
+For icons implementation trough the whole app we have two approaches. The first one is by adding 
+custom svg files to the project and the second one by using the library `react-native-vector-icons`.
+
+The first approach with custom svg icon files is recommended when the icons are specially made for a certain project. 
+To use this methode refer to the [Assets](https://infinum.com/handbook/books/frontend/assets) section in this handbook.
+
+## react-native-vector-icons
+
+Otherwise if the project contains of standard icons that are available in icon sets like `Material icons, FontAwsome, AntDesign` etc.
+it is better to use the `react-native-vector-icons`. To install the library follow the instructions on [https://github.com/oblador/react-native-vector-icons](https://github.com/oblador/react-native-vector-icons).
+
+You don't need to import all available icon sets. It's enough to import just the icon set/s that you use inside the project.
+Also the library provides you with a `Icon` component that receives 3 main props, `name` for the icon name, `size` for the icons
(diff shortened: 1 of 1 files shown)


React Native/Icons.md
```diff
@@ -0,0 +1,45 @@
...
+
+The first approach with custom svg icon files is recommended when the icons are specially made for a certain project. 
+To use this methode refer to the [Assets](https://infinum.com/handbook/books/frontend/assets) section in this handbook.
+
+## react-native-vector-icons
+
+Otherwise if the project contains of standard icons that are available in icon sets like `Material icons, FontAwsome, AntDesign` etc.
+it is better to use the `react-native-vector-icons`. To install the library follow the instructions on [https://github.com/oblador/react-native-vector-icons](https://github.com/oblador/react-native-vector-icons).
+
+You don't need to import all available icon sets. It's enough to import just the icon set/s that you use inside the project.
```
andreicek, line 12:
```suggestion
You should not import all available icon sets. It's enough to import just the icon set/s that are used (see example below).
```

React Native/Icons.md
```diff
@@ -0,0 +1,45 @@
...
+Also the library provides you with a `Icon` component that receives 3 main props, `name` for the icon name, `size` for the icons
+size in dpi's (default is 12) and `color` for the color of the icon (default is inherited).
+For other styling props and static methodes for the `Icon` component you can refer to the previouse linked documentation.
```
andreicek, line 15:
```suggestion
For other styling props and static methodes for the `Icon` component you can refer to the previously linked documentation.
```

andreicek (changes requested):
Thanks @SusanjRaul! Awesome guidelines. I've left just a few comments on wording.

Since this is our guidelines we use "should" instead of "could".

React Native/Icons.md
```diff
@@ -0,0 +1,44 @@
+There are two methods on how we use and implement icons. The first one is by adding
```
veksi, line 1: What it means `custom svg icon files`?
Maybe this should be more explained, since as far as I am aware just importing svg icons doesn't work. You have to either use `react-native-svg` library or custom transformer to use svg icons as is or "custom".

SusanjRaul: I can change this to something like:

The first one is by adding custom svg icon files to the project and using a library called react-native-svg ...

In the next paragraph is a reference to another handbook section (which is in review currently) that explains in detail how to use this method.


What do you think about this change?

PR #133, opened 2021-08-17
feat(angular): angular-specific datx related information section

Erbenos:
Just some Angular-specific stuff I feel like is worth noting since there is already a plethora of other DatX resources available either be it official docs or the DatX article so I have felt as if I am duplicating content. If you feel differently, please let me know.

Also was unable to check it against the handbook app as I was running into some issues setting it up.

### DIFF
--- a/Angular/Angular guidelines and best practices/DatX data store.md
+++ b/Angular/Angular guidelines and best practices/DatX data store.md
@@ -23,7 +23,7 @@ Since we didn't include mobx (because it doesn't play nice with RxJS), we need t
 import '@datx/core/disable-mobx';
 ```
 
-Then we need to overwrite mobx path so that it can be resolved by datx, albeit to an **empty** file:
+Next, we need to overwrite mobx path so that it can be resolved by datx, albeit to an **empty** file:
 
 ```ts
 // /tsconfig.json
@@ -39,7 +39,7 @@ Then we need to overwrite mobx path so that it can be resolved by datx, albeit t
 }
 ```
 
-Lastly we bypass `@datx/core/disable-mobx` related import warning by adding its non ECMAScript module dependency to a whitelist:
+Lastly, we bypass `@datx/core/disable-mobx` related import warning by adding its non ECMAScript module dependency to a whitelist:
 
 ```ts
 // angular.json
@@ -60,11 +60,11 @@ Lastly we bypass `@datx/core/disable-mobx` related import warning by adding its 
 
 ## Defining models and their relationships
 
-This topic is heavily described in the resources above, so please, refer to them. By convention you can put all models in `src/app/models` directory and collections to `src/app/collections` of your Angular project. Once you are finished describing all _Resource Objects_ you can move to the next section, just remember that you will most likely want to use _jsonapiAngular_ so that you can work with observables when calling async methods on models.
+This topic is heavily described in the resources above, please refer to them. By convention, you can put all models in `src/app/models` directory and collections to `src/app/collections` of your Angular project. Once you are finished describing all _Resource Objects_ you can move to the next section, just remember that you will most likely want to use _jsonapiAngular_ so that you can work with observables when calling async methods on models.
 
 ## Store setup and injection
 
-To create and provide single instance of Collection accross entire Angular app, you can create custom DI token eg. APP_COLLECTION, see [InjectionToken docs](https://angular.io/api/core/InjectionToken) and then provide said token in root module, like so:
+To create and provide a single instance of Collection accross entire Angular app, you can create custom DI token eg. `APP_COLLECTION`, see [InjectionToken docs](https://angular.io/api/core/InjectionToken) and then provide said token in root module, like so:
 
 ```ts
 // injection-tokens.ts
@@ -101,7 +101,7 @@ import { AppCollection } from '<path-to-collection-definition>';
 export class AppModule {}
 ```
 
-Then whenever you want to work with the collection from within Angular's DI container, you can simply use the token above to inject the collection instance.
+Whenever you want to work with the collection from within Angular's DI container, you can simply use the token above to inject the collection instance.
 
 ```ts
 // example.service.ts
@@ -119,7 +119,7 @@ export class ExampleService {
 }
 ```
 
-When testing consumers of APP_COLLECTION, just inject the collection under the same token again as with any other dependency:
+When testing consumers of APP_COLLECTION, just inject the collection under the same token again, as with any other dependency:
 
 ```ts
 // example.service.spec.ts
@@ -151,11 +151,11 @@ describe('ExampleService', () => {
 
 ## Configuring DatX HTTP calls interception
 
-By default, DatX uses Fetch API when invoking HTTP calls, but this means that it also bypass **HttpClient** from `@angular/common/http` and therefore any registered interception logic or similar kind of middleware. Luckily there is a way around that, see [official documentation guide](https://datx.dev/docs/jsonapi-angular/base-fetch).
+By default, DatX uses Fetch API when invoking HTTP calls, but this means that it also bypasses **HttpClient** from `@angular/common/http` and therefore any registered interception logic or similar kind of middleware. Luckily there is a way around that, see [official documentation guide](https://datx.dev/docs/jsonapi-angular/base-fetch).
 
 ## Abstracting entity related login into separate service
 
-You might want to abstract away a little of the tediousness of working with DatX and/or bound some special logic to a given entity. Since many of the Collection's method require an Entity type as a parameter you can create separate service that will hide this small implementation detail for you. One way to do that would be to extend an abstract class similar to this one:
+You might want to abstract away a bit of the tediousness of working with DatX and/or bind some special logic to a given entity. Since many of the Collection's methods require an Entity type as a parameter you can create separate service that will hide this small implementation detail for you. One way to do that would be to extend an abstract class similar to this one:
 
 ```ts
 // collection.service.ts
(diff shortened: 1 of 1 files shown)


Angular/Angular guidelines and best practices/DatX data store.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Why use DatX
+
+Apart from providing possible single source of truth for any shared data with a basic querying and referencing support, it also provides options to extend any entity with mixins such as [withActions](https://datx.dev/docs/mixins/with-actions), [withMeta](https://datx.dev/docs/mixins/with-meta), [withPatches](https://datx.dev/docs/mixins/with-patches), which enhance entities with usefull utility functions, or network related ones like [withNetwork](https://datx.dev/docs/mixins/with-network), [jsonapi](https://datx.dev/docs/mixins/jsonapi) or, useful in Angular's case, [jsonapiAngular](https://datx.dev/docs/mixins/jsonapi-angular). Especially with [jsonapi](https://datx.dev/docs/mixins/jsonapi) and/or [jsonapiAngular](https://datx.dev/docs/mixins/jsonapi-angular), interacting with [JSONAPI standard](https://jsonapi.org/) compliat backend can be much easier then building requests manually and more managable than building API related code with some sort of generator.
+
+## Dependencies
+Let's start by installing all DatX related dependencies we will need:
+```
+npm i --save @datx/core @datx/jsonapi @datx/jsonapi-angular mobx
```
DarkoKukovec, line 14: With angular, we usually avoid using mobx since it's an overhead and it doesn't play nice (at least not out of the box) with rxjs.
Because of this, we need the noop.js file you might have noticed in the project (and referenced in angular.json) and the `@datx/core/disable-mobx` import in the app.

Angular/Angular guidelines and best practices/DatX data store.md
```diff
@@ -0,0 +1,206 @@
...
+		const model = this.create(rawModel);
+		return this.save(model);
+	}
+
+	public findAll(): Array<TModel> {
+		return this.collection.findAll<TModel>(this.ctor);
+	}
+
+	public getAllModels(options?: IRequestOptions): Observable<Array<TModel>> {
+		return this.getMany({
```
DarkoKukovec, line 147: What does this exactly do? Seems to me like it might be something project specific?

Angular/Angular guidelines and best practices/DatX data store.md
```diff
@@ -0,0 +1,206 @@
...
+## Configuring DatX HTTP calls interception
+
+By default, DatX uses Fetch API when invoking HTTP calls, but this means that it also bypass **HttpClient** from `@angular/common/http` and therefore any registered interception logic or similar kind of middleware. Luckily there is a way around that, see [official documentation guide](https://datx.dev/docs/jsonapi-angular/base-fetch).
+
+## Abstracting entity related login into separate service
+
+You might want to abstract away a little of tediousness of working with DatX and/or bound some special logic to a given entity. Since many of the Collection's method require an Entity type as a parameter you can create separate service that will hide this small implementation detail for you. One way to do that would be to extend an abstract class similar to this one:
+
+```ts
+// collection.service.ts
```
DarkoKukovec, line 114: Looking now at this, would it make sense to move it to `@datx/jsonapi-angular` instead of c/p-ing it in every project? @fvoska @kristijanbambir

Angular/Angular guidelines and best practices/DatX data store.md
```diff
@@ -0,0 +1,206 @@
...
+
+	it('should be created', () => {
+		expect(service).toBeTruthy();
+	});
+});
+```
+
+## Configuring DatX HTTP calls interception
+
+By default, DatX uses Fetch API when invoking HTTP calls, but this means that it also bypass **HttpClient** from `@angular/common/http` and therefore any registered interception logic or similar kind of middleware. Luckily there is a way around that, see [official documentation guide](https://datx.dev/docs/jsonapi-angular/base-fetch).
```
DarkoKukovec, line 107: Also, should we also move base fetch into the lib to avoid copy/paste? Was there a reason why we didn't do it like this right away?

Angular/Angular guidelines and best practices/DatX data store.md
Erbenos: Expanded on that with a3f71e1

Angular/Angular guidelines and best practices/DatX data store.md
Erbenos: Yes, it's optional, wording fits "You might want to abstract away a little of the tediousness of working with DatX and/or bound some special logic to a given entity.". Do you want me to remove it?

fvoska:
Can you please add some more details about testing as well? Things about `CollectionTestingSerice` and how to create test doubles for model-specific service. Things like this:
```ts
@Injectable()
export class AuthoritiesTestingService
	extends CollectionTestingService<Authority>
	implements ExtractPublic<AuthoritiesService> {
	protected ctor = Authority;

	constructor(@Inject(APP_COLLECTION) protected readonly collection: AppCollection) {
		super(collection);
	}
}
```

Angular/Angular guidelines and best practices/DatX data store.md
DarkoKukovec: No, it can stay, but I'm not sure what does the `getAllModels` represent - I think we're missing a better example if we want to show custom logic for a single entity.

Erbenos: Added comment explaining that in 7dc1020

Erbenos:
Expanded on testing of collection service and model-specific services in 7dc1020

Angular/Angular guidelines and best practices/DatX data store.md
fvoska: I think that would be OK, but then `@datx/jsonapi-angular` will need to add angular as a dependency (peer + dev?), which it currently does not have (that is kinda nice).

Angular/Angular guidelines and best practices/DatX data store.md
fvoska: I think that for now we can leave it in each app separately since we probably won't need it in datx v3 anyway so it will become obsolete?

Angular/Angular guidelines and best practices/DatX data store.md
```diff
@@ -0,0 +1,467 @@
...
+      useValue: new AppCollection();
+    }
+    ...
+  ],
+  bootstrap: [AppComponent],
+}
+export class AppModule {}
+```
+
+Then whenever you want to work with the collection from within Angular's DI container, you can simply use the token above to inject the collection instance.
```
fvoska, line 104:
```suggestion
Whenever you want to work with the collection from within Angular's DI container, you can simply use the token above to inject the collection instance:
```

Angular/Angular guidelines and best practices/DatX data store.md
```diff
@@ -0,0 +1,467 @@
...
+  providedIn: 'root',
+})
+export class ExampleService {
+  constructor(
+    @Inject(APP_COLLECTION) protected readonly collection: AppCollection // app.module.ts scoped instance of AppCollection
+  ) {}
+}
+```
+
+When testing consumers of APP_COLLECTION, just inject the collection under the same token again as with any other dependency:
```
fvoska, line 122:
```suggestion
When testing consumers of APP_COLLECTION, just inject the collection under the same token again, as with any other dependency:
```

Angular/Angular guidelines and best practices/DatX data store.md
```diff
@@ -0,0 +1,467 @@
...
+});
+```
+
+## Configuring DatX HTTP calls interception
+
+By default, DatX uses Fetch API when invoking HTTP calls, but this means that it also bypass **HttpClient** from `@angular/common/http` and therefore any registered interception logic or similar kind of middleware. Luckily there is a way around that, see [official documentation guide](https://datx.dev/docs/jsonapi-angular/base-fetch).
+
+## Abstracting entity related login into separate service
+
+You might want to abstract away a little of the tediousness of working with DatX and/or bound some special logic to a given entity. Since many of the Collection's method require an Entity type as a parameter you can create separate service that will hide this small implementation detail for you. One way to do that would be to extend an abstract class similar to this one:
```
fvoska, line 158:
```suggestion
You might want to abstract away a bit of the tediousness of working with DatX and/or bind some special logic to a given entity. Since many of the Collection's methods require an Entity type as a parameter you can create separate service that will hide this small implementation detail for you. One way to do that would be to extend an abstract class similar to this one:
```

Angular/Angular guidelines and best practices/DatX data store.md
```diff
@@ -0,0 +1,467 @@
...
+	...
+})
+export class EntityListComponent {
+	public readonly entities$ = this.entity.getAllModels();
+
+	constructor(private readonly entity: EntityService) {}
+}
+```
+
+When testing these abstractions, it advisable to create doubles, which replace any methods that would query service with an implementation which check in memory collection like this example **CollectionTestingService**:
```
fvoska, line 292:
```suggestion
When testing these abstractions, it is advisable to create test doubles that replace any methods that would make API calls with an implementation which works with in memory collection, like in this example **CollectionTestingService**:
```

fvoska (changes requested):
Good job 👏 , this will be really useful when starting new Angualr projects that use DatX!

Angular/Angular guidelines and best practices/DatX data store.md
DarkoKukovec: I don't think adding angular as a dependency is a problem, we can use `*` or something equally broad in the peerDependency versions so we don't complicate the Angular updates.

Erbenos:
Addressed fvoska's grammar comments with 4f49ef

DarkoKukovec (approved):

fvoska (approved):

PR #134, opened 2021-08-25
Datx provider recipe

kristian240:
Add `useDatx` and `DatxProvier` recepies. Recipe is API agnostic, however, there is one section for JSON:API.

### DIFF
--- a/React/Recipes/Session Handling.md
+++ b/React/Recipes/Session Handling.md
@@ -113,16 +113,22 @@ export function useSession({
     ...config,
   });
 
+  const callbackRefs = useRef({ onLoginSuccess, onLoginError, onLogoutSuccess, onLogoutError });
+
+  useEffect(() => {
+    callbackRefs.current = { onLoginSuccess, onLoginError, onLogoutSuccess, onLogoutError };
+  });
+
   const login = useCallback(
     async (attributes) => {
       const session = createSession(datx, attributes).then(
         (session) => {
-          onLoginSuccess?.(session);
+          callbackRefs.current.onLoginSuccess?.(session);
 
           return session;
         },
         (error) => {
-          onLoginError?.(error);
+          callbackRefs.current.onLoginError?.(error);
 
           return Promise.reject(error);
         }
@@ -130,21 +136,21 @@ export function useSession({
 
       return state.mutate(session, false);
     },
-    [datx, onLoginError, onLoginSuccess, state]
+    [datx, state]
   );
 
   const logout = useCallback(async () => {
     const session = deleteSession(datx).then(
-      () => onLogoutSuccess?.(),
+      () => callbackRefs.current.onLogoutSuccess?.(),
       (error) => {
-        onLogoutError?.(error);
+        callbackRefs.current.onLogoutError?.(error);
 
         return Promise.reject(error);
       }
     );
 
     return state.mutate(session, false);
-  }, [datx, onLogoutError, onLogoutSuccess, state]);
+  }, [datx, state]);
 
   return { login, logout, state };
 }


React/Recipes/Datx Store Provider.md
```diff
@@ -0,0 +1,100 @@
+## datx
+
+> [datx.dev - A mobx data store](https://datx.dev/)
+>
+> DatX is an opinionated JS/TS data store. It features support for simple property definition, references to other models, and first-class TypeScript support.
+>
+> To find out more about Datx, check out the [official documentation](https://datx.dev/).
```
DarkoKukovec, line 7:
```suggestion
> To find out more about DatX, check out the [official documentation](https://datx.dev/).
```
Just to be consistant

React/Recipes/Session Handling.md
```diff
@@ -1,67 +1,168 @@
...
+
+```ts
+import { SessionModel } from 'models/SessionModel';
+
+/**
+  Creates and returns a session from an endpoint.
+  User model is included in the response.
+*/
+export async function createSession(datx, loginData) {
+  return fetch(SESSION_API_ENDPOINT, {
```
DarkoKukovec, line 41: Sessions are maybe a wrong example for it, but it would be useful to see how this fetcher would look like with jsonapi

React/Recipes/Session Handling.md
```diff
@@ -100,15 +204,17 @@ const SomePrivatePage = () => {
...
+  const { state } = useSession();
 
   const session = state.data;
 
   return {session ? <PrivateContent /> : <Loading />}
 }
 ```
 
-Although this might look like a simple page redirect, render will occur and might hurt app performance. Once the page is loaded and the component is mounted, `state.data` will be `undefined`. Once request to read a session, `state.data` or `state.error` will be defined, and re-render will occur. This means that all child components will be re-rendered, too, which can potentially be an expensive operation. To optimize this, we'll create a component that will handle the redirect.
+Although this might look like a simple page redirect, it's not really optimized - rerender will occur and might hurt performance.
```
DarkoKukovec, line 215: I'm not sure if it's a good idea to show an example that is not good enough, at least not as a first example.

DarkoKukovec (commented):
Looks quite good now. I like the details and step by step explanations of the code.

DarkoKukovec (approved):

React/Recipes/Session Handling.md
isBatak: How can we demonstrate this pitfall in a better way? I often see this in our projects, and I don't think our devs knows about this pitfall.

isBatak (changes requested):

React/Recipes/Datx Store Provider.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+Once this is set up, and if we assume that all steps defined in this document are done, we can start fetching. For this example, we'll create a component that will fetch `todos` from the API and render them on the page.
+
+Before we create a component, we will create a fetcher.
+
+```tsx
+export const getTodos = (datx) => {
+  return fetch('/todos')
+    .then((res) => res.json())
```
isBatak, line 150: I would use async/await here

React/Recipes/Session Handling.md
```diff
@@ -1,67 +1,168 @@
...
+  Creates and returns a session from an endpoint.
+  User model is included in the response.
+*/
+export async function createSession(datx, loginData) {
+  return fetch(SESSION_API_ENDPOINT, {
+    method: 'POST',
+    body: loginData,
+    // ...rest of the options
+  })
+    .then((res) => res.json())
```
isBatak, line 46: Also, can we use async/await here?

isBatak (changes requested):

React/Recipes/Datx Store Provider.md
```diff
@@ -0,0 +1,188 @@
...
+
+interface IDatxProviderProps {
+  client: Client;
+}
+
+export const DatxProvider: FC<IDatxProviderProps> = ({ client, children }) => (
+  <DatxContext.Provider value={client}>
+    <SWRConfig
+      value={{
+        compare: dequal,
```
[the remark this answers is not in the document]
kristian240: In terms of a core datx, is it necessary to have a custom compare function? Looking at the [dequal docs](https://github.com/lukeed/dequal#dequallite), it should be able to compare class instances. I'm correct or not?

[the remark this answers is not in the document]
isBatak: I'm not sure anymore, but we had some problem with default compare before on some project. 
@DarkoKukovec do you remember?

Also you are importing some custom `compare` function on line 51 which is not used anywhere. What's with that?

React/Recipes/Datx Store Provider.md
```diff
@@ -0,0 +1,192 @@
...
+> For frontend network layer we will use [SWR](https://swr.vercel.app/) library created by Vercel.
+
+```tsx
+// DatxProvider.tsx
+
+import { createContext, FC } from 'react';
+import { SWRConfig } from 'swr';
+import { dequal } from 'dequal/lite';
+
+import { compare } from 'utils/compare';
```
isBatak, line 51: Isn't this a custom compare function that fixes some DatX detection problems...

isBatak (changes requested):

React/Recipes/Datx Store Provider.md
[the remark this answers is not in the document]
DarkoKukovec: Not sure, I'm not aware of any issues, but I wasn't that involved in this part, so I might have missed it.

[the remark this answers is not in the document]
isBatak: @DarkoKukovec I think you were involved in this conversation with @danijelbuhin or @mamihelj09 or someone. 
Problem happens when API returns new data and we already have previous data, SWR will try to compare previous value and new one. But as MobX is mutable, previous state in the SWR cache will also be updated and compare function will have the two identical objects and change won't be detected. 
I think something like that happened, it's because of MobX mutability.

Something like this:
```
const data = new MobxObservableObject({ name: 'test' }); // I forgot the syntax sorry 

// SWR cache
const cache = [{ key: 'some/url', value: data }];

// update happens
mutate(() => data.save());

// SWR compare will take the previous data in the cache and compare it with the new, and if there is change it will trigger component update
compare(cache[0].value, data);
```
As data is mutable observable object it will be updated in the SWR cache immediately after save is resolved, so new state and the old state will be the same...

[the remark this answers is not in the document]
DarkoKukovec: Yes, that's why the snapshots should be used - they should be immutable

PR #140, opened 2021-11-03
Update Angular file structure

fvoska:
The main change here is the possibility to have pages folder inside some other page, in order to better organize routing when there are multiple levels. This helps a lot when there are many levels of routing.

There is one point of discussion around naming, regarding `-index` components. Previoulsy we would probably name these `-container`, We need _some_ postfix because otherwise there are components with duplicate names. For example, we need a naming distinction between `user-details(1)` component that is loaded via routing (eagerly) and `user-details(2)` component that actually renders user details. `user-details(1)` component renders some sort of a header/footer, maybe fetches some data from the API and renders `user-details(2)` with that data. After a discussion with @kristijanbambir and taking a look at some other projects, `-index` seems to be a good option. Thusly, we named `user-details(1)` `user-index` and `user-details(2)` is simply `user-details`. I would like to hear some other opinions.

### DIFF
--- a/Angular/Angular guidelines and best practices/File and module organization and naming.md
+++ b/Angular/Angular guidelines and best practices/File and module organization and naming.md
@@ -29,11 +29,11 @@ As an example, please consider the following routing setup:
 
 - `/` - homepage, redirects to `/posts`
 - `/users` - list of users (renders a list of `user-avatar` components)
-- `/users/:id/view` - user details view (renders `user-avatar` and a list of `post-overview` components)
+- `/users/:id` - user details view (renders `user-avatar` and a list of `post-overview` components)
 - `/posts` - list of posts (renders a list of `post-overview` and `user-avatar` components)
 - `/posts/new` - new post form (renders `post-form`)
+- `/posts/:id` - post details view (renders `post-overview`, `user-avatar` and `post-content` components)
 - `/posts/:id/edit` - edit post form (renders `post-form`)
-- `/posts/:id/view` - post details view (renders `post-overview`, `user-avatar` and `post-content` components)
 
 Directory structure for such an app should look like this:
 
@@ -66,7 +66,7 @@ src/app/
 │   │   │       ├── user-index.component.ts # renders some header/footer and users-details component
 │   │   │       ├── user-index-routing.module.ts # eagerly-loads user-index.component on /
 │   │   │       └── user-index.module.ts # imports user-index-routing.module
-│   │   ├── users-routing.module.ts # lazy-loads users-list.module on /, user-details.module on /:id/view
+│   │   ├── users-routing.module.ts # lazy-loads users-list.module on /, user-details.module on /:id
 │   │   └── users.module.ts # imports users-routing.module
 │   └── posts/
 │       ├── components/ # these components are used only in posts feature/domain module
@@ -95,7 +95,7 @@ src/app/
 │       │       ├── edit-post.component.ts # renders some header/footer and post-form component
 │       │       ├── edit-post-routing.module.ts # eagerly-loads edit-post.component on /
 │       │       └── edit-post.module.ts # imports edit-post-routing.module
-│       ├── posts-routing.module.ts # lazy-loads posts-index.module on /, new-post.module on /new, edit-post.module on /:id/edit, post-index.module on /:id/view
+│       ├── posts-routing.module.ts # lazy-loads posts-index.module on /, new-post.module on /new, edit-post.module on /:id/edit, post-index.module on /:id
 │       └── posts.module.ts # imports posts-routing.module
 ├── app-routing.module.ts # lazy-loads users.module on /users, posts.module on /posts
 └── app.module.ts # imports app-routing.module
--- a/Angular/Angular guidelines and best practices/File and module organization and naming.md
+++ b/Angular/Angular guidelines and best practices/File and module organization and naming.md
@@ -70,8 +70,7 @@ src/app/
 │   │   └── users.module.ts # imports users-routing.module
 │   └── posts/
 │       ├── components/ # these components are used only in posts feature/domain module
-│       │   ├── post-form/ # form used for both editing and creation
-│       │   └── post-content/ # renders post content
+│       │   └── post-form/ # form used for both editing and creation
 │       ├── services/
 │       │   └── social-sharing/ # service that is used only inside posts module
 │       ├── pages/
@@ -83,7 +82,7 @@ src/app/
 │       │   │   └── posts-index.module.ts # imports posts-index-routing.module
 │       │   ├── post-index/
 │       │   │   ├── components/
-│       │   │   │   └── post-details/ # renders post-overview, user-avatar and post-content components
+│       │   │   │   └── post-details/ # renders post-overview and user-avatar components and renders the post content
 │       │   │   ├── post-index.component.ts # renders some header/footer and posts-details component
 │       │   │   ├── post-index-routing.module.ts # eagerly-loads post-index.component on /
 │       │   │   └── post-index.module.ts # imports post-index-routing.module
--- a/Angular/Angular guidelines and best practices/File and module organization and naming.md
+++ b/Angular/Angular guidelines and best practices/File and module organization and naming.md
@@ -23,61 +23,87 @@ src/app/services/post/
 
 ## Pages and shared components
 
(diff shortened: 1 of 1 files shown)


Angular/Angular guidelines and best practices/File and module organization and naming.md
```diff
@@ -15,53 +15,113 @@ Try to keep the code grouped by domain. For example, if you have a `Post` class,
...
-Container components that are loaded directly via routing (by either `component` or `loadChildren`) should be placed in the `src/app/pages` directory.
+Components that are loaded directly via routing (by either `component` or `loadChildren`) should be placed in the `src/app/pages` directory. If a component is used in multiple places, it should be extracted to the nearest common ancestor's components directory. If there are multiple levels of routing, the main `src/app` directory structure can be mirrored under some specific page directory. In other words, you can have components, services, pages, etc. directories inside some specific page directory. This helps separate the codebase on a feature/domain-basis.
 
-## Shared components
+As an example, please consider the following routing setup:
 
-If a component is used in multiple places, it should be extracted to the nearest common ancestor's components directory. Here are some examples:
+- `/` - homepage, redirects to `/posts`
+- `/users` - list of users (renders a list of `user-avatar` components)
+- `/users/:id/view` - user details view (renders `user-avatar` and a list of `post-overview` components)
```
Erbenos, line 32: Do we really want to force the "view" suffix here?

Angular/Angular guidelines and best practices/File and module organization and naming.md
```diff
@@ -15,53 +15,113 @@ Try to keep the code grouped by domain. For example, if you have a `Post` class,
...
+│   │   │       │   └── user-details/ # renders user-avatar component, additional user info and a list of user's posts (post-overview)
+│   │   │       ├── user-index.component.ts # renders some header/footer and users-details component
+│   │   │       ├── user-index-routing.module.ts # eagerly-loads user-index.component on /
+│   │   │       └── user-index.module.ts # imports user-index-routing.module
+│   │   ├── users-routing.module.ts # lazy-loads users-list.module on /, user-details.module on /:id/view
+│   │   └── users.module.ts # imports users-routing.module
+│   └── posts/
+│       ├── components/ # these components are used only in posts feature/domain module
+│       │   ├── post-form/ # form used for both editing and creation
+│       │   └── post-content/ # renders post content
```
fvoska, line 74: I now noticed that `post-content` is used only in post-details component so it doesn't really need to be this high up in the folder hierarchy. I will think about moving it deeper in the hierarchy or adding another place where it will be used

Angular/Angular guidelines and best practices/File and module organization and naming.md
fvoska: I am flexible on this. With REST APIs we have different HTTP verbs so the routes are usually just `/users/:id` and then you have GET, POST, etc. I am not sure what's the most common thing for frontend apps, I've seen many different approaches (`users/:id`, `users/:id/view`, `users/:id/details`, `users/view/:id`, etc.)

Erbenos: My view (hehe) on this is that there is no reason to specify we are viewing a single resource with additional suffix in the URL, :id of a resource is enough to clarify that, with additional verbs/nouns such as "edit" implying a different action a resource, in this case, one can understand /users/:id:/view as us browsing "a view of users :id", where "a view" != "a detail of a resource".

safo6m:
> There is one point of discussion around naming, regarding `-index` components. Previoulsy we would probably name these `-container`, We need _some_ postfix because otherwise there are components with duplicate names. For example, we need a naming distinction between `user-details(1)` component that is loaded via routing (eagerly) and `user-details(2)` component that actually renders user details. `user-details(1)` component renders some sort of a header/footer, maybe fetches some data from the API and renders `user-details(2)` with that data. After a discussion with @kristijanbambir and taking a look at some other projects, `-index` seems to be a good option. Thusly, we named `user-details(1)` `user-index` and `user-details(2)` is simply `user-details`. I would like to hear some other opinions.

I'm used to `-container` naming but other than that I don't have any objections to the `-index` naming.
@fvoska You mentioned that some projects are using the `-index` convention, could you list some of those projects (I'm just curious)?

Angular/Angular guidelines and best practices/File and module organization and naming.md
```diff
@@ -15,53 +15,113 @@ Try to keep the code grouped by domain. For example, if you have a `Post` class,
...
+│       ├── services/
+│       │   └── social-sharing/ # service that is used only inside posts module
+│       ├── pages/
+│       │   ├── posts-index/
+│       │   │   ├── components/
+│       │   │   │   └── posts-list/ # renders post-overview and user-avatar component for each post
+│       │   │   ├── posts-index.component.ts # renders some header/footer and posts-list component
+│       │   │   ├── posts-index-routing.module.ts # eagerly-loads posts-index.component on /
+│       │   │   └── posts-index.module.ts # imports posts-index-routing.module
```
thabalija, line 83: Should `posts-intex.testing.module.ts` and `posts-index.mock.component.ts` also be here?

fvoska: No, because these components are loaded via routing (not via template of some other component) so you do not need to import mock/testing versions of those components in tests for some other componen.

Angular/Angular guidelines and best practices/File and module organization and naming.md
fvoska: > > There is one point of discussion around naming, regarding `-index` components. Previoulsy we would probably name these `-container`, We need _some_ postfix because otherwise there are components with duplicate names. For example, we need a naming distinction between `user-details(1)` component that is loaded via routing (eagerly) and `user-details(2)` component that actually renders user details. `user-details(1)` component renders some sort of a header/footer, maybe fetches some data from the API and renders `user-details(2)` with that data. After a discussion with @kristijanbambir and taking a look at some other projects, `-index` seems to be a good option. Thusly, we named `user-details(1)` `user-index` and `user-details(2)` is simply `user-details`. I would like to hear some other opinions.
> 
> I'm used to `-container` naming but other than that I don't have any objections to the `-index` naming. @fvoska You mentioned that some projects are using the `-index` convention, could you list some of those projects (I'm just curious)?

[Polyglot](https://github.com/infinum/js-polyglot-angular/tree/master/src/app/pages/projects/pages) used component name index, although it was not as a postfix but as the complete name.

Angular/Angular guidelines and best practices/File and module organization and naming.md
fvoska: Sold, I will drop `/view`

ivan-mitric:
My two cents regarding the topic (without the need to start a larger discussion):

From the start, the idea to split the files by (business) domain was much closer to me than splitting them to pages, components, services, etc.

For example, I would have an object folder, with all these other subfolders inside:

category
 |- components
 |- pages
 |- services
post
 |- components
 |- pages
 |- services

It just makes sense to me that these files are sorted like this, each folder can have its own (router) module if needed.

I guess that with this approach the list of components would be shorter and it would be easier to navigate through it.

I realise that this would be a big change, I would just like to hear opinions.

kristijanbambir:
> My two cents regarding the topic (without the need to start a larger discussion):
> 
> From the start, the idea to split the files by (business) domain was much closer to me than splitting them to pages, components, services, etc.
> 
> For example, I would have an object folder, with all these other subfolders inside:
> 
> category |- components |- pages |- services post |- components |- pages |- services
> 
> It just makes sense to me that these files are sorted like this, each folder can have its own (router) module if needed.
> 
> I guess that with this approach the list of components would be shorter and it would be easier to navigate through it.
> 
> I realise that this would be a big change, I would just like to hear opinions.

@ivan-mitric what about components and services that are used both in category and post domains, where would they go?

In this proposal we are still splitting by domain, but that happens within the pages folder. A subpage (essentially a domain) can have the same folder structure as the application root (pipes, components, services, …).

fvoska:
@ivan-mitric what you are proposing is almost exactly the same as is proposed with the changes in this PR, the only difference is that these domain-directories need to live _somewhere_ and the proposed solution is that they live in `pages` because it is almost always the case that different application domains will also be separated by top-level routing. At least I think that should be the case. There might be some exceptions but we can handle those as they come.

Erbenos (approved):

PR #142, opened 2021-11-04
Add info about test host component pattern

fvoska:
Expanded the chapter about testing components that use `OnPush` CD with host component example and ngneat/spectator as an alternative option.


Angular/Angular guidelines and best practices/Testing.md
```diff
@@ -531,6 +531,96 @@ beforeEach(async(() => {
 }));
 ```
 
+This is, arguably, not the best solution but it might be the simplest solution and good enough for some cases.
```
kristijanbambir, line 534:
```suggestion
This is, arguably, not the best solution, but it might be the simplest, and good enough for some cases.
```

Angular/Angular guidelines and best practices/Testing.md
```diff
@@ -531,6 +531,96 @@ beforeEach(async(() => {
...
+
+### Alternative approach with a host/wrapper component
+
+Changing change detection strategy from OnPush to Default in tests is not ideal because that could theoretically make the component behave differently when comparing application runtime where it is using OnPush and tests where it is using Default CD.
```
kristijanbambir, line 538:
```suggestion
Changing the change detection strategy from OnPush to Default in tests is not ideal because that could theoretically make the component behave differently when comparing the application runtime, where it is using OnPush, and tests, where it is using the Default CD.
```

Angular/Angular guidelines and best practices/Testing.md
```diff
@@ -531,6 +531,96 @@ beforeEach(async(() => {
...
+
+An alternative approach is to wrap the component you want to test into another component that is declared only in the TestBed and is used only for interacting with the component you want to test via inputs and outputs.
+
+Fixture is created for the host component and `fixture.componentInstance` will not be an instance of the component you want to test, it will be an instance of the host component. If you need the instance of the component you want to test, you can get it after the fixture is initialized, via `debugElement.query` or by adding a `ViewChild` in the host component.
+
+Host component pattern starts to shine when you need to change some input value and trigger CD - you can simply change the public property value on the host component and call `fixture.detectChanges()`. That will trigger `ngOnChanges` and re-render the child component, even if it is using OnPush CD. If you didn't use host component, you would have to manually assign property values and call `ngOnChanges` with the correct `changes` object. That can be tedious to write, or you could even forget to call `ngOnChanges` and then wonder why the tests are not working as expected.
```
kristijanbambir, line 544:
```suggestion
The host component pattern starts to shine when you need to change some input value and trigger CD - you can simply change the public property value on the host component and call `fixture.detectChanges()`. That will trigger `ngOnChanges` and re-render the child component, even if it is using the OnPush CD. If you didn't use the host component, you would have to manually assign property values and call `ngOnChanges` with the correct `changes` object. That can be tedious to write, and you could even forget to call `ngOnChanges` and then wonder why the tests are not working as expected.
```

Angular/Angular guidelines and best practices/Testing.md
```diff
@@ -531,6 +531,96 @@ beforeEach(async(() => {
...
+
+The downside of using a host component is that it is a bit boilerplate-y - you will have to bind potentially many inputs and outputs between the host and child components.
```
kristijanbambir, line 546:
```suggestion
The downside of using a host component is that it is a bit boilerplate-y - you will potentially have to bind many inputs and outputs between the host and child components.
```

Angular/Angular guidelines and best practices/Testing.md
```diff
@@ -531,6 +531,96 @@ beforeEach(async(() => {
...
+    [value]="value"
+    (valueChange)="onValueChange()
+  >
+  </counter>
+  `
+})
+class CounterHostComponent {
+  public value: number = 0; // Notice that this doesn't have to be an @Input
+
+  public onValueChange(newValue: number) { }
```
kristijanbambir, line 583:
```suggestion
  public onValueChange(newValue: number): void { }
```

Angular/Angular guidelines and best practices/Testing.md
```diff
@@ -531,6 +531,96 @@ beforeEach(async(() => {
...
+      providers: [...],
+    }).compileComponents();
+  })
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(CounterHostComponent);
+    fixture.detectChanges();
+
+    hostComponent = fixture.componentInstance;
+    component = fixture.debugElement.query(By.css('counter')).componentInstance;
```
kristijanbambir, line 604: I would instead use a ViewChild in the CounterHostComponent to create the component instance variable and then use it here.

Angular/Angular guidelines and best practices/Testing.md
```diff
@@ -531,6 +531,96 @@ beforeEach(async(() => {
...
+    counterButton.click();
+
+    expect(valueChangeHandlerSpy).toHaveBeenCalledTimes(1);
+  });
+});
+```
+
+### Another alternative, ngneat/spectator
+
+If you are willing to add a 3rd party library to the project, [ngneat/spectator](https://github.com/ngneat/spectator) has some useful helpers for setting inputs and triggering CD, even for OnPush components. However, it will change the way you set up tests so it is up to you if you want to go this route or not.
```
kristijanbambir, line 622: How do I know am I willing to add it? 🙂  The idea of the Handbook is to be a guide for what to use in projects. I would  either recommend to use this package, or omit it completely. Since I'm hearing about this package for the first time, I would omit it. Otherwise, it should state when should I use this package, and when not.

kristijanbambir (changes requested):

Angular/Angular guidelines and best practices/Testing.md
fvoska: Ok I will remove it.

Angular/Angular guidelines and best practices/Testing.md
fvoska: I will change it to ViewChild, but either solution would be fine with me

kristijanbambir (approved):

PR #143, opened 2021-11-19
Angular assets urls

fvoska:
This is a woozy about how the URLs/paths to assets should be defined in order to make it work in different cases (``, css url(), different base paths).

TODO:

- [x] Highlight the conclusions for all the different cases

You can preview the chapter on staging: https://beta.infinum.com/handbook/frontend/angular/angular-guidelines-and-best-practices/assets-and-caching#paths-to-assets


Angular/Angular guidelines and best practices/Assets & caching.md
```diff
@@ -0,0 +1,172 @@
...
+
+## JS and CSS chunks
+
+JS and CSS chunks that are linked to from the `index.html` file are fingerprinted and because of that they can be cached indefinitely. If we create a new application build, the `index.html` file will be re-validated and the client will load new chunks.
+
+**Rule #2**: Cache fingerprinted JS and CSS files with a long expiration time!
+
+## Static assets
+
+The tricky part with serving Angular applications is caching the static assets files. Some frameworks, depending on Webpack configuration and the way that the assets are used, fingerprint all the images and other static assets. In Angular, static assets like images and custom fonts are placed in the `src/assets/` directory. This directory gets copied without any modifications to the final application build directory - `dist/[project-name]/assets/`.
```
kristijanbambir, line 107:
```suggestion
The tricky part with serving Angular applications is caching the static assets files. Some frameworks, depending on the Webpack configuration and the way that assets are used, fingerprint all images and other static assets. In Angular, static assets, like images and custom fonts, are placed in the `src/assets/` directory. This directory gets copied without any modifications to the final application build directory - `dist/[project-name]/assets/`.
```

Angular/Angular guidelines and best practices/Assets & caching.md
```diff
@@ -0,0 +1,172 @@
...
+
+The files from the `assets/` directory do not get fingerprinted. Because of this, it is important to set up the correct caching mechanism for serving these files or else the client might be using stale files. There are two solutions:
```
kristijanbambir, line 109:
```suggestion
Files from the `assets/` directory do not get fingerprinted. Because of this, it is important to set up the correct caching mechanism for serving these files or else the client might be using stale files. There are two solutions:
```

Angular/Angular guidelines and best practices/Assets & caching.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Cache re-validation implementation
+
+Depending on how the application files are served, implementation details for setting up HTTP caching will be a bit different.
+
+We will cover some of the common HTTP serving technologies, including static file serving and Server-side rendering.
+
+### Static file serving
+
+If the application files are served statically, the static file server must be configured to serve and cache assets files correctly.
```
kristijanbambir, line 134:
```suggestion
If application files are served statically, the static file server must be configured to serve and cache assets files correctly.
```

Angular/Angular guidelines and best practices/Assets & caching.md
```diff
@@ -0,0 +1,172 @@
...
+
+#### Nginx example
+
+The Nginx configuration file for the site should be updated to enable cache re-validation for the `assets/` directory:
```
kristijanbambir, line 138: We are handling this on CEERIS at the moment with the `cache-control: no-cache` header and our devops team already added that to the nginx [template configuration](https://github.com/infinum/devops-ansible/blob/master/roles/nginx-vhost/templates/single-page-angular.j2) for Angular SPA applications. ETags are turned on by default in nginx. no-cache will compare ETags and download the new versions if values don't match. I think this should be written here.

Angular/Angular guidelines and best practices/Assets & caching.md
```diff
@@ -0,0 +1,172 @@
...
+            add_header Surrogate-Control "public, max-age=0";
+            add_header Cache-Control "public, max-age=0";
+            add_header Cache-Control must-revalidate;
+        }
+    }
+```
+
+### Server-side rendering
+
+If the application uses Server-side rendering, then the caching should be handled by the Node.js server.
```
kristijanbambir, line 159:
```suggestion
If the application uses server-side rendering, then caching should be handled by the Node.js server.
```

Angular/Angular guidelines and best practices/Assets & caching.md
```diff
@@ -0,0 +1,172 @@
...
+When using Angular Universal, `server.ts` includes a piece of code in charge of serving the application client build files (this includes `index.html`, JS and CSS chunks, and files from `assets/`). This piece of code should be updated to force cache re-validation:
+
+```typescript
+server.get('*.*', express.static(distFolder, {
+  maxAge: '0',
+  etag: true,
+}));
+```
+
+This change will actually force re-validation for `index.html`, JS and CSS chunks as well, not only the `assets/` directory. This is ok, you could even disable output hashing of chunk files.
```
kristijanbambir, line 170: >you could even disable output hashing of chunk files

Could or should?

kristijanbambir (changes requested):

Angular/Angular guidelines and best practices/Assets & caching.md
fvoska: If all chunks are re-validated, you could theoretically turn off output hashing, but there is no harm in leaving it on.

Angular/Angular guidelines and best practices/Assets & caching.md
```diff
@@ -0,0 +1,172 @@
...
+| `/` | `/assets/logo.png` | Absolute | `/assets/logo.png` | <span style="color:green;">✔</span> | - |
+| `/web/` | `/web/assets/logo.png` | Absolute | `/web/assets/logo.png` | <span style="color:green;">✔</span> | - |
+
+Notes:
+
+- Same rules will apply to any asset loading via `url()` (e.g. fonts).
+- `d808b08ce47c5d0c53cd` is example of a hash, different values will appear in reality.
+- In certain cases, the logo image will be present two times in the final bundle:
+  - `/dist/logo.d808b08ce47c5d0c53cd.png` and `/dist/assets/logo.png`
+- `assets/logo.png` + `<base href="/web/">` would work and it would load the image from `https://app.com/web/assets/logo.png` if SCSS transpiled, but it does not. If you set this URL manually in devtools, you will notice that it does indeed work and takes into account `base` `href` value.
```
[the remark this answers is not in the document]
fvoska: Actually here i used `transpiled` as a verb, not adjective, so I will revert the change about `if SCSS transpiled`. What is meant here is that SCSS transpilation fails.

Angular/Angular guidelines and best practices/Assets & caching.md
kristijanbambir: Yes, but in the context of the handbook, we should be explicit about what to do.

fvoska: Ok, I will remove that sentence

Angular/Angular guidelines and best practices/Assets & caching.md
fvoska: I updated both the nginx and node examples to set no-cache header.

Angular/Angular guidelines and best practices/Assets & caching.md
```diff
@@ -0,0 +1,172 @@
...
+| `/web/` | `../../../../../../../assets/logo.svg` | `/assets/logo.svg` | `https://app.com/web/assets/logo.svg` | <span style="color:red;">⨯</span> |
+| `/web/` | `./assets/logo.svg` | `/web/assets/logo.svg` | `https://app.com/web/assets/logo.svg` | <span style="color:green;">✔</span> |
+| `/web/` | `/assets/logo.svg` | `/assets/logo.svg` | `https://app.com/web/assets/logo.svg` | <span style="color:red;">⨯</span> |
+| `/web/` | `assets/logo.svg` | `/web/assets/logo.svg` | `https://app.com/web/assets/logo.svg` | <span style="color:green;">✔</span> |
+
+Notes:
+
+- All the `<img>` `src` paths listed above are relative paths, but paths that start with `/` will ignore `base` `href` and resolve relative to the domain ([w3c docs](https://www.w3schools.com/tags/att_img_src.asp)). All other relative paths will take into account `base` `href` when resolving the path.
+
+In summary, you must use `./assets/…` or `assets/…` when defining paths for assets loading in both HTML and CSS in order to make it work in all the cases. We recommend using `./assets/…`.
```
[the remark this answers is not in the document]
fvoska: I have added a Conclusion sub-heading for each of the sections.

PR #159, opened 2022-04-06
#323 Document NextJS routing

mislavmiocevic:


React/NextJS routing.md
```diff
@@ -0,0 +1,428 @@
...
+
+```tsx
+// SSR pages/article/[articleId].tsx
+
+import { useRouter } from 'next/router';
+
+const ArticlePage = ({ articleId }) => {
+    const { query } = useRouter()
+
+    if (!query.articleId) {
```
isBatak, line 80: You don't need this if because you are getting `articleId` form the props, so you don't need `useRouter` in this example

React/NextJS routing.md
```diff
@@ -0,0 +1,428 @@
...
+
+```tsx
+// SSG pages/article/[articleId].tsx
+
+import { useRouter } from 'next/router';
+
+const ArticlePage = ({ articleId }) => {
+    const { query } = useRouter()
+
+    if (!query.articleId) {
```
isBatak, line 109: Same as above

React/NextJS routing.md
```diff
@@ -0,0 +1,428 @@
...
+You can change the base path if your application needs to be under a sub-path of a domain.
+To do that, change the `basePath` property in the `next.config.js` file.
+
+
+## 404 pages
+
+Routing to non-existing pages is working out of the box. 
+Only if there is a `SSG` or a `SSR` then you need to handle it manually. To implement that, you should return: 
+```
+notFound: true
```
isBatak, line 402: Could you add the whole example with `getSetverSideProps`

isBatak (changes requested):

React/NextJS routing.md
mislavmiocevic: Again, same as my above comment.

React/NextJS routing.md
mislavmiocevic: Sure

React/NextJS routing.md
isBatak: it's not a real world example, also it's a bad practice, so I'll remove the code

mislavmiocevic: Will do.

React/NextJS routing.md
mislavmiocevic: Done
https://github.com/infinum/frontend-handbook/pull/159/commits/eca4b6642691d8e1f5b4e38ebde662fa6c98cfb9

isBatak:
@mislavmiocevic should this file be under Recipes section? 
it is at least in this PR https://github.com/infinum/frontend-handbook-private/pull/52/files

React/NextJS routing.md
```diff
@@ -0,0 +1,455 @@
...
+
+export const getStaticPaths = async () => {
+    return {
+        paths: [],
+        fallback: 'blocking'
+    }
+};
+
+export const getStaticProps = async (context) => {
+    const articleId = context.params.articleId;
```
isBatak, line 407: You are not setting `context.params` in `getStaticPaths`

mislavmiocevic:
@isBatak I honestly do not know and I am confused with the current structure in the repo. 

I wanted to create the `next js` folder and move all NextJS related things there.  What do you think about it? And we can move others later in some other PR. Also who decides the structure so we can invite to join us in the discussion.

isBatak:
@mislavmiocevic lets put it under Recipes, we can do a restructure later

React/NextJS routing.md
mislavmiocevic: getStaticPaths does not require any paths so you can have empty array. Here the topic is routing, and this part is related to static generation and data fetching mechanisms.

isBatak:
Some styles are broken https://beta.infinum.com/handbook/frontend/react/recipes/nextjs-routing

isBatak:
I'm not actually sure if we need this chapter in the handbook. All of this is documented in the Next.js documentation, with this we are just creating an overhead because we need to maintain this file if something is changed in the Next.js docs.

Is there something we can extract from this document and put it in a recipe, like some common mistake that our teammates are doing and we want to suggest a better approach or avoid that habit?

isBatak:
I'm not actually sure if we need this chapter in the handbook. All of this is documented in the Next.js documentation, with this we are just creating an overhead because we need to maintain this file if something is changed in the Next.js docs.

Is there something we can extract from this document and put it in a recipe, like some common mistake that our teammates are doing and we want to suggest a better approach or avoid that habit?

isBatak:
Should we then split the important things "knowhow" that are not in the docs into a smaller recipe chapters?

mislavmiocevic:
> Should we then split the important things "knowhow" that are not in the docs into a smaller recipe chapters?

Okay, let's do it like that then. I will see what can be and how splitted. If you have some ideas, feel free to share.

mislavmiocevic:
Closign this PR because we will take part from it and create independent recipes.

PR #171, opened 2022-06-06
Add Angular SSR content

zlatko-stulic:


Angular/Server-side rendering (SSR)/SEO.md
```diff
@@ -0,0 +1,362 @@
...
+Twitter will use Open Graph meta tags if twitter ones are not set. You don't need to set twitter meta tags except if you want a different contest for the twitter cards.
+
+ * `<meta name="twitter:card" content="summary  | summary_large_image">`
+
+### Header Tags
+You can use header tags to change font sizes and signify information hierarchy on a page.
+The heading elements go from H1 to H6. H1 is the largest and most important level, and H6 is the smallest and least important.
+
+## SEO service
+In this example here we will cover service example for the SEO:
```
fvoska, line 58:
```suggestion
The following example shows how you might implement a service that makes it easier to set various tags:
```

Angular/Server-side rendering (SSR)/SEO.md
```diff
@@ -0,0 +1,362 @@
...
+		this.metaService.removeTag('property="og:image:height"');
+		this.metaService.removeTag('property="og:image:type"');
+		this.metaService.removeTag('property="og:type"');
+		this.metaService.removeTag('property="og:locale"');
+		this.metaService.removeTag('name="twitter:card"');
+	}
+}
+```
+
+For using this SEO service it is best to create `guard` which will call this service before accessing the page. You can define route data and read it in the `guard`:
```
fvoska, line 254:
```suggestion
For using this SEO service it is best to create a `guard` which will call this service before accessing the page. You can define route data and read it in the `guard`:
```

Angular/Server-side rendering (SSR)/SEO.md
```diff
@@ -0,0 +1,362 @@
...
+    canActivate: [SeoContentGuard],
+    runGuardsAndResolvers: 'always',
+}
+```
+
+This is example if your application is using translation, so you will provide translation keys.
+
+`runGuardsAndResolvers: 'always'` is used to be sure that guard will be fired on navigation back.
+
+To be able to use translations there is function `createTranslatedSeoContent` in the service which is using the `TranslocoService` to get translations. In the last step lets crate the guard:
```
fvoska, line 272:
```suggestion
To translate the content, we use `createTranslatedSeoContent` and `TranslocoService`. The last step is to create the guard:
```

Angular/Server-side rendering (SSR)/Transfer state.md
```diff
@@ -0,0 +1,45 @@
+When we use Angular Universal, the API that delivers the content is hit twice. First when the server is rendering the page and second, when the application is bootstrapped. This causes latency issues and a bad user experience as the screen usually flickers when this happens.
```
fvoska, line 1:
```suggestion
When we use Angular Universal, the API that delivers the content is hit twice. First time is when the server is rendering the page and the second time is when the application is bootstrapped in the browser. This causes latency issues and a bad user experience as the screen usually flickers when this happens. Not to mention that this is not optimal.
```

Angular/Server-side rendering (SSR)/Transfer state.md
```diff
@@ -0,0 +1,45 @@
...
+
+We can use `TransferState` service to send information from the server to the client, which avoids making duplicate API calls.
+
+## TransferHttpCacheModule
+
+`TransferHttpCacheModule` installs a http interceptor that avoids duplicate `HttpClient` GET requests on the client, for requests that were already made when the application was rendered on the server side.
```
fvoska, line 7:
```suggestion
`TransferHttpCacheModule` installs an HTTP interceptor that avoids duplicate `HttpClient` GET requests on the client, for requests that were already made when the application was rendered on the server side.
```

Angular/Server-side rendering (SSR)/Transfer state.md
```diff
@@ -0,0 +1,45 @@
...
+@NgModule({
+	imports: [AppModule, ServerModule, ServerTransferStateModule],
+	bootstrap: [AppComponent],
+})
+export class AppServerModule {}
+```
+
+### Verification
+
+You can test that `TransferState` has been configured successfully if you go the Network section in your browser and see that all initial GET request are no more listed. We can see also that script `<script id="serverApp-state" type="application/json">...</script>` is injected.
```
fvoska, line 45:
```suggestion
You can test that `TransferState` has been configured successfully if you go the Network section in your browser and see that all initial GET request are no longer being made. You might also notice that a script element was added to the initial document response - `<script id="serverApp-state" type="application/json">...</script>`, including all the transfered data.
```

fvoska (changes requested):
Good job writing these documents 👏

Most of my comments are wording suggestions, please take a look at them

Angular/Server-side rendering (SSR)/Robots and sitemap.md
```diff
@@ -0,0 +1,114 @@
...
+
+ * [apicache](https://www.npmjs.com/package/apicache) for caching the response
+ * [sitemap](https://www.npmjs.com/package/sitemap) for generating sitemap.xml file
+
+```typescript
+server.get('/sitemap.xml', cache('1 hour'), (req, res) => {
+		res.header('Content-Type', 'application/xml');
+
+		// If there is no API_URL environment variable return server error and stop creating sitemap
+		if (!apiUrl) {
```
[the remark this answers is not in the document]
zlatko-stulic: I will remove this part as we had translations in both projects, so it can be a bit confusing for other people where route translations are used. We could add route translation as a separate chapter in the handbook for Angular.

fvoska (approved):

PR #172, opened 2022-06-17
Add icon folder to React project structure

kristian240:
I see different locations where devs place the icons folder. With this PR I propose that the folder should live in `src/components/icons`.

### DIFF
--- a/React/Project structure.md
+++ b/React/Project structure.md
@@ -6,6 +6,7 @@ When adding UI components, you should be able to group them in two root domains:
 1. `core` - primitives, low level components
 2. `shared` - components that are shared all across the app
 3. `features` - root folder for components based on a specific feature (could be scoped by page or island)
+3. `icons` - SVG icons used throughout the application
 
 Folder naming rules:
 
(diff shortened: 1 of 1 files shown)


isBatak:
Icons are by the definition `shared` components (highly reusable). Should we keep them under `shared` folder? 
If we start extracting domains (in this case domain "icon") then we could do it for the layouts domain too.
What bothers me about 'layouts` is the rule that shared component should not use any other shared component, only the core primitives (Chakra UI), but `layouts` usually has to import navigation and footer shared components.

We need to make a decision here, do we want to extract `icons` and `layouts` out of the shared folder, or keep them both in.

kristian240:
I always thought of icons as assets, not really as components, even though they are `svg` component. Previously, they were held, in most of the projects, in `src/assets/icons` in `.svg` file. Now with chakra, we have the option to keep them as components. That's why my first instinct is to place them in `src/components/icons`.

However, I can see how this extraction of the domain can cause the confusion and might tempt the devs to extract other domains as well, like `layouts` as you mentioned. 

I agree they can be placed in `src/components/shared/icons` instead but nevertheless we should add this to handbook as there is many ways a developer is storing them in the project - me and you e.g.

I'm updating the PR.

PR #173, opened 2022-06-20
Add Testing passed props section

kristian240:
https://app.productive.io/1-infinum/tasks/2876257

### DIFF
--- a/React/Testing best practices.md
+++ b/React/Testing best practices.md
@@ -571,11 +571,27 @@ describe("User Page", () => {
 
 ## Testing passed props
 
-When testing the user interactions or trying to see if a component is called with a certain property you can wrap a component in a jest function.
+A quick how to on using Jest to check if correct props are passed to a child component.
 
-In the example below, the component with a button and on click it will open a modal. We want to test if the `Modal` is open or not by checking if the `Modal` component got `isOpen={true}`
+This example is purely to show how to verify that a React components props are passed in a Jest unit test. There are two components, a `MyModal` and a `Modal` from Chakra UI library. The `MyModal` renders `Modal` and `Button`, and the 'open' state is handled with `useDisclosure` hook. The goal is to check if `Modal` component gets the `isOpen={true}` prop when the user clicks the button.
 
 ```tsx
+// MyModal.tsx
+export const MyModal: FC<ButtonProps> = (props) => {
+  const { isOpen, onOpen, onClose } = useDisclosure();
+
+  return (
+    <>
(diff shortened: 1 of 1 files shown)


React/Testing best practices.md
```diff
@@ -569,6 +569,43 @@ describe("User Page", () => {
 });
 ```
 
+## Testing passed props
+
+When testing the user interactions or trying to see if a component is called with a certain property you can wrap a component in a jest function.
```
isBatak, line 574: maybe we can simplify the intro... code speaks more than words :D
```suggestion
A quick how to on using Jest to check if correct props are passed to a child component.
```

React/Testing best practices.md
```diff
@@ -569,6 +569,43 @@ describe("User Page", () => {
...
+			const { Modal: OriginalModal } = jest.requireActual('@chakra-ui/react');
+
+			return <OriginalModal {...props} />;
+		}),
+	};
+});
+
+describe("MyModal", () => {
+	it("should open on click", () => {
+		render(<MyModal />);
```
isBatak, line 596: Do we need to show an example of `MyModal` component?

isBatak: with useDisclosure and stuff

React/Testing best practices.md
```diff
@@ -569,6 +569,43 @@ describe("User Page", () => {
 });
 ```
 
+## Testing passed props
+
+When testing the user interactions or trying to see if a component is called with a certain property you can wrap a component in a jest function.
+
+In the example below, the component with a button and on click it will open a modal. We want to test if the `Modal` is open or not by checking if the `Modal` component got `isOpen={true}`
```
isBatak, line 576:
```suggestion
This example is purely to show how to verify that a React components props are passed in a Jest unit test. There are two components, a `MyModal` and a `Modal` from Chakra UI library. The `MyModal` renders `Modal` and `Button`, and 'open' state is handled with `useDisclosure` hook. The goal is to check if `Modal` component got the `isOpen={true}` prop when user clicks the button.
```

isBatak (changes requested):

React/Testing best practices.md
kristian240: I added it!

React/Testing best practices.md
kristian240: Changed

React/Testing best practices.md
kristian240: Changed

isBatak (approved):

PR #180, opened 2022-10-13
Add naming convention table

isBatak:


React/Naming Cheatsheet.md
```diff
@@ -0,0 +1,22 @@
+## Naming Cheatsheet
```
kristian240, line 1: We don't need a root heading, handbook will use name of the file as the first heading. We can remove this and change all other levels to `level - 1`. E.g., `### Naming Components -> ## Naming Components`.

kristian240 (changes requested):
You are adding a new chapter, you also need to update the `chapters.yaml` file.

fvoska:
I think we could move this to some framework-agnostic part, this makes sense for Angular as well

isBatak:
> I think we could move this to some framework-agnostic part, this makes sense for Angular as well
@fvoska I was thinking the same thing, but...
What should I do with the **Note** part? 
Do you have something similar to polymorphic components in Angular?

Naming Cheat Sheet.md
```diff
@@ -0,0 +1,29 @@
...
+External link to community maintained cheatsheet:  
+[Naming cheatsheet](https://github.com/kettanaito/naming-cheatsheet/blob/main/README.md)
+
+## Naming Components
+
+### HC/LC/b/p/s Pattern
+There is a useful pattern to follow when naming functions:
+
+```
+high context (HC)? + low context? (LC)? + base + part? + suffix?
```
isBatak, line 16:
```suggestion
high context (HC)? + low context (LC)? + base + part? + suffix?
```

fvoska (approved):

fvoska:
@isBatak this has been opened for a while. I am fine with the proposal, so if there are no additional comments from others, I think we can merge.

isBatak:
@fvoska thx, let me just check it on staging one more time

PR #185, opened 2022-12-14
Expand observables postfixing section

danipavic:
This PR expands the `Postfix observables with $` section with the information/conclusions from this [discussion](https://github.com/infinum/frontend-handbook/discussions/181).

### DIFF
--- a/Angular/Angular guidelines and best practices/Formatting, naming and best practices.md
+++ b/Angular/Angular guidelines and best practices/Formatting, naming and best practices.md
@@ -586,6 +586,23 @@ const observable$: Observable;
 const mySubject$: Subject;
 ```
 
+The `$` suffix is here to indicate and let you know it is something you can subscribe to, but it does not let you know if you can change the observable in case it some type of a `Subject`, ie. `ReplaySubject` or `BehaviorSubject`. This naming convention goes along with the `no-exposed-subjects` RxJS ESLint rule. The aforementioned rule will cause an error in case you have publicly exposed` Subject`. The reasoning behind this is that subjects should never be publicly exposed unless converted to an observable using some pipeable operators or for an example, `asObservable` function.
+
+```typescript
+class ExampleComponent {
+  private _isLoading$ = new BehaviorSubject(false); // for internal use for "state management"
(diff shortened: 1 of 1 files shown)


Angular/Angular guidelines and best practices/Formatting, naming and best practices.md
```diff
@@ -586,6 +586,23 @@ const observable$: Observable;
 const mySubject$: Subject;
 ```
 
+The `$` suffix is here to indicate and let you know it is something you can subscribe to, but it does not let you know if you can change the observable in case it some type of a `Subject`, ie. `ReplaySubject` or `BehaviorSubject`. This naming convention goes along with the `no-exposed-subjects` RxJS ESLint rule. The aforementioned rule will cause an error in case you have publicly exposed` Subject`. The reasoning behind this is that subjects should never be publicly exposed unless converted to an observable using some pipeable operators or for an example, `asObservable` function.
```
thabalija, line 589:
```suggestion
The `$` suffix is here to indicate and let you know it is something you can subscribe to, but it does not let you know if you can change the observable in case it is some type of a `Subject`, ie. `ReplaySubject` or `BehaviorSubject`. This naming convention goes along with the `no-exposed-subjects` RxJS ESLint rule. The aforementioned rule will cause an error in case you have publicly exposed` Subject`. The reasoning behind this is that subjects should never be publicly exposed unless converted to an observable using some pipeable operators or for example, `asObservable` function.
```

Angular/Angular guidelines and best practices/Formatting, naming and best practices.md
```diff
@@ -586,6 +586,23 @@ const observable$: Observable;
...
+  // to be called from some methods, for example during data fetching
+  private updateLoadingState(state) {
+    this._isLoading$.next(state);
+  }
+}
+```
+
+The potential downside of this convention/pattern is the name separation, in case of a naming conflict, the private member name should be prefixed with the `_`.
+
+
```
thabalija, line 605:
```suggestion

```

thabalija (changes requested):

thabalija (approved):

fvoska (changes requested):

fvoska (approved):

PR #186, opened 2022-12-19
Managing node versions

danipavic:
This PR adds a chapter about managing Node versions. I currently have a dilemma if another section should be added stating what preferred/go-to version manager should be used team/project-wise. I wouldn't force anyone to use this or that, but having a default package seems like a good idea.

### DIFF
--- a/Node/Managing Node-NPM versions.md
+++ b/Node/Managing Node-NPM versions.md
@@ -2,34 +2,59 @@
 
 ### Node release schedule
 
-Node.js has defined and predictable release plan. Each node version goes through 3 phases: current, active(LTS) and maintenance. Be aware that odd-numbered versions will not go through active and maintenance phases.
+Node.js has defined and predictable release plan. Depending on the Node release version number, the Node release can go through 3 different phases: current, active(LTS) and maintenance. Be aware that odd-numbered versions will not go through active and maintenance phases, while even numbered will go through them all.
 
 #### Current phase
 
-Incorporates most non-major changes that end up on `nodejs/node` main branch. New major releases are branched out every six months. Even-numbered versions are scheduled for release in April, while odd ones are scheduled for October.
+Releases within this phase incorporate most non-major changes that end up on `nodejs/node` main branch. New major releases are branched out every six months. Even-numbered versions are scheduled for release in April, while odd ones are scheduled for October.
 
 #### Active Long Term Support phase
 
-Versions in this phase include new features, bug fixes and updates that have been audited and determined to be stable for the release. With each odd-numbered release, previous even-numbered release will transition to LTS phase. Versions in this phase will receive active support for 12 months after they entered LTS phase.
+Releases within this phase include new features, bug fixes and updates that have been audited and determined to be stable for the release. With each odd-numbered release, previous even-numbered release will transition to LTS phase. Versions in this phase will receive active support for 12 months after they entered LTS phase.
 
 **You should always aim to use latest LTS version available on your project**
 
 #### Maintenance phase
 
-Versions in this phase will receive critical bug fixes and security updates for 18 months after it enters this phase.
+Releases within this phase will receive critical bug fixes and security updates for 18 months in case of LTS(even-numbered) and 2 months in case of regular(odd-numbered) releases after entering this phase.
 
 ### Ensuring same version of Node is being used on all environments
 
-It is very important to keep the same versions of Node across the environments. When you update the version of Node in your project, you should take appropriate actions that the new version is also propagated/updated on build/production servers by notifying people responsible for those environments. By not running the same versions you could run into unexpected problems with failing builds, incorrect dependencies etc.
+It is very important to keep the same version of Node across the environments. When you update the version of Node in your project, you should take appropriate actions that the new version is also propagated/updated on build/production servers by notifying people responsible for those environments. By not running the same versions you could run into unexpected problems with failing builds, incorrect dependencies etc.
 
 ### Managing versions
 
-There are couple of ways for managing versions of Node. I will list couple of options and generally, all options are fine as long as you are proficient at installing, updating and switching Node versions using that method. These probably aren't all available version managers on the internet, but are probably most popular.
+There are couple of ways for managing versions of Node. This section lists couple of options and generally, all options are fine as long as you are proficient at installing, updating and switching Node versions using that method. These probably aren't all available version managers on the internet, but are probably most popular.
+
+#### n
+
+[n](https://github.com/tj/n) is both powerful and easy to use version manager for Node and should be preferred method for managing Node versions.
+You can create a `.n-node-version`, `.node-version` or `.nvmrc` file containing a Node version number. Running `n auto` will look for version specified in one of those files in that order and switch to it.
+
+```bash
+#Install n by running script
+npm install -g n
+
+#Install latest Node version
+n latest
+
+#Install specific version of Node by specifying version or use an alias
+n <version>
+n lts
+
+#Execute to check Node version currently in use and available versions
+n
+
+#Uninstall Node version(this does not affect cached versions)
+n uninstall <version>
+
+#Uninstall cached Node version
+n rm <version>
+```
 
 #### nvm - Node version manager
 
-[Node version manager](https://github.com/nvm-sh/nvm) is both powerful and easy to use and should be preferred method for managing Node versions.
-You can create a `.nvmrc` file containing a Node version number. Running `use`, `install`, `which` etc. commands will use the version specified in the file if none was provided in the command line. When using `nvm` and installing global packages, those packages will only be global for that specific version. eg. You currently use Node 16 and install `polyglot-cli`, if you switch to v18, the package won't be available there until you install it.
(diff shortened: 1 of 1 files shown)


Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,127 @@
+## Node version management
+
+### Node release schedule
+
+Node.js has defined and predictable release plan. Each node version goes through 3 phases: current, active(LTS) and maintenance. Be aware that odd-numbered versions will not go through active and maintenance phases.
```
DarkoKukovec, line 5: Kind of misleading saying that every node version will go trough a LTS phase.

Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,127 @@
...
+
+#### Current phase
+
+Incorporates most non-major changes that end up on `nodejs/node` main branch. New major releases are branched out every six months. Even-numbered versions are scheduled for release in April, while odd ones are scheduled for October.
```
DarkoKukovec, line 9:
```suggestion
Releases within this phase incorporate most non-major changes that end up on `nodejs/node` main branch. New major releases are branched out every six months. Even-numbered versions are scheduled for release in April, while odd ones are scheduled for October.
```

Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,127 @@
...
+
+#### Active Long Term Support phase
+
+Versions in this phase include new features, bug fixes and updates that have been audited and determined to be stable for the release. With each odd-numbered release, previous even-numbered release will transition to LTS phase. Versions in this phase will receive active support for 12 months after they entered LTS phase.
```
DarkoKukovec, line 13:
```suggestion
Releases within this phase include new features, bug fixes and updates that have been audited and determined to be stable for the release. With each odd-numbered release, previous even-numbered release will transition to LTS phase. Versions in this phase will receive active support for 12 months after they entered LTS phase.
```

Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,127 @@
...
+
+**You should always aim to use latest LTS version available on your project**
+
+#### Maintenance phase
+
+Versions in this phase will receive critical bug fixes and security updates for 18 months after it enters this phase.
```
DarkoKukovec, line 19:
```suggestion
Releases within this phase will receive critical bug fixes and security updates for 18 months after it enters this phase.
```

DarkoKukovec: Actually, it's 18 moths for LTS maintenance, but only 2 months for regular maintenance (odd versions).

Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Ensuring same version of Node is being used on all environments
+
+It is very important to keep the same versions of Node across the environments. When you update the version of Node in your project, you should take appropriate actions that the new version is also propagated/updated on build/production servers by notifying people responsible for those environments. By not running the same versions you could run into unexpected problems with failing builds, incorrect dependencies etc.
```
DarkoKukovec, line 23:
```suggestion
It is very important to keep the same version of Node across the environments. When you update the version of Node in your project, you should take appropriate actions that the new version is also propagated/updated on build/production servers by notifying people responsible for those environments. By not running the same versions you could run into unexpected problems with failing builds, incorrect dependencies etc.
```

Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Managing versions
+
+There are couple of ways for managing versions of Node. I will list couple of options and generally, all options are fine as long as you are proficient at installing, updating and switching Node versions using that method. These probably aren't all available version managers on the internet, but are probably most popular.
```
DarkoKukovec, line 27: The phrasing here shouldn't be in first person.

Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,127 @@
...
+
+#### nvm - Node version manager
+
+[Node version manager](https://github.com/nvm-sh/nvm) is both powerful and easy to use and should be preferred method for managing Node versions.
```
DarkoKukovec, line 31: We talked on the leads meeting the other day and the conclusion is that nvm is not an ideal solution because it doesn't use the more standardised `.node-version` file.

Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,127 @@
...
+#Uninstall Node version(this does not affect cached versions)
+n uninstall <version>
+
+#Uninstall cached Node version
+n rm <version>
+```
+
+#### Using homebrew
+
+For this method you need to have [homebrew](https://brew.sh/) installed on your Mac(which you probably do). There might be additional steps required when switching versions using `brew` as switching is not supported, so you might need to uninstall and unlink the old version, and install and link new one.
```
DarkoKukovec, line 87: I think we should activelly discourage people of doing that because they will have issues at some point because they'll need multiple versions.

Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,127 @@
...
+
+#Install specific version of node
+brew install node@<version>
+
+#Update Node
+brew upgrade node
+
+#Uninstall Node
+brew uninstall node
+```
```
DarkoKukovec, line 101: [nodenv](https://github.com/nodenv/nodenv) is probably also worth mentioning since it is one of the more popular options.

Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,127 @@
...
+#### avn - Automatic Version Switching
+
+[avn](https://github.com/wbyoung/avn) is a package for automatic version switching of Node using one of your Node version manager package. Currently supported are `nvm`, `n` and `nodebrew`. When you `cd` into a directory containing `.node-version` file, `avn` will automatically detect the change and use version manager of your choice to switch to the specified version.
+
+```bash
+#Install the packages
+npm install -g avn avn-nvm avn-n
+avn setup
+```
+
```
DarkoKukovec, line 116: Given the [`.node-version` support](https://github.com/shadowspawn/node-version-usage) and avn support, I'm guessing the good recommendation would be to use `n`?

Node/Managing Node-NPM versions.md
danipavic: Decided to keep the section but added a note about discouraging the usage of this method.

Node/Managing Node-NPM versions.md
```diff
@@ -56,35 +81,38 @@ nvm use default
 nvm uninstall <version>
 ```
 
-#### n
+#### nvm - Node version manager
```
DarkoKukovec, line 84: Two times?

Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,126 @@
...
+
+Releases within this phase include new features, bug fixes and updates that have been audited and determined to be stable for the release. With each odd-numbered release, previous even-numbered release will transition to LTS phase. Versions in this phase will receive active support for 12 months after they entered LTS phase.
+
+**You should always aim to use latest LTS version available on your project**
+
+#### Maintenance phase
+
+Releases within this phase will receive critical bug fixes and security updates for 18 months in case of LTS(even-numbered) and 2 months in case of regular(odd-numbered) releases after entering this phase.
+
+### Ensuring same version of Node is being used on all environments
```
fvoska, line 21: I would expend this chapter a bit. If we are using SSR (either Next.js or Angular Universal), then you also have to make sure that you use the same node version on both build server during the build step AND on production server when running the app.

Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,126 @@
...
+n rm <version>
+```
+
+#### nvm - Node version manager
+
+[Node version manager](https://github.com/nvm-sh/nvm) is a version manager similar to `n` and a good alternative. You can create a `.nvmrc` file containing a Node version number. Running `use`, `install`, `which` etc. commands will use the version specified in the file if none was provided in the command line. When using `nvm` and installing global packages, those packages will only be global for that specific version. eg. You currently use Node 16 and install `polyglot-cli`, if you switch to v18, the package won't be available there until you install it.
+
+```bash
+#Install NVM by running script
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.2/install.sh | bash
```
fvoska, line 61: I would avoid writing installation instructions for any of the mentioned node version management tools, as these steps are likely to change in the future and we would have to update our handbook. You even have nvm version in this command. I think it will be sufficient to just link to the official website for the tool and instruct the reader to follow the instructions there.

Node/Managing Node-NPM versions.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+#### Specifying version in package.json
+
+You can specify the version of node that your stuff works on by providing Node versions in package.json. If you don't specify this, it means any version will do.
+
+```json
+{
+  "engines": {
+    "node": ">= 18.12"
```
fvoska, line 123: We should also write down npm version here, as minor NPM versions can also be breaking changes (*cough* [`npm v8.4.1`](https://infinum.slack.com/archives/C01MJ2HDA05/p1655806084190259) *cough*)

Node/Managing Node-NPM versions.md
fvoska: I think that in this chapter, we should make a recommendation which tool to use. Base on all of the tools listed here, `n` seems like the best option to me.

fvoska (changes requested):

fvoska:
lgtm+

fvoska (approved):

PR #190, opened 2023-02-02
Add canMatch guard and auto-unwrap sections

danipavic:
This PR adds `canMatch` guard and auto-unwrap sections in the handbook. Initially, I only added a section about `canMatch` but since both is kinda related to the router I added both in the same PR.


Angular/Angular guidelines and best practices/Formatting, naming and best practices.md
```diff
@@ -1048,6 +1048,31 @@ Some notes:
 - We wrapped everything in one `ng-container` to avoid calling the pipe twice
 - Be careful in case that your pipe can return a valid falsy value, as in that case the `*ngIf` will not render the content; in such case you might consider using a [custom `*ngLet`](https://github.com/ngrx-utils/ngrx-utils#nglet-directive) structural directive
 
+## Auto unwrap default exports when lazy loading
+
+With Angular v15 it is possible to leverage default exports to shorten the syntax when lazy loading module or a standalone component. Note that the CLI still generates modules and components without the default export so you will need to add that by yourself.
```
Erbenos, line 1053: a module

Angular/Angular guidelines and best practices/Formatting, naming and best practices.md
```diff
@@ -1048,6 +1048,31 @@ Some notes:
...
+
+```typescript
```
Erbenos, line 1055: Just FYI I think "\`ts\`" works too

Angular/Angular guidelines and best practices/Formatting, naming and best practices.md
```diff
@@ -1048,6 +1048,31 @@ Some notes:
...
+@Component({...})
```
Erbenos, line 1056: It only makes sense to use loadComponent with standalone components, make it clear here. You even mention it in text, but this component isn't standalone

Angular/Angular guidelines and best practices/Formatting, naming and best practices.md
```diff
@@ -1048,6 +1048,31 @@ Some notes:
...
+// With default export
+{
+  path: 'lazy',
+  loadComponent: () => import('./lazy-file'),
+}
+```
+
+## Favouring canMatch guard
+
+The [canMatch](https://angular.io/api/router/CanMatch) is the new type of guard introduced in Angular `v14.1`. It should be the preferred guard to use over `canActivate` or `canLoad`  which is deprecated in `v15`. `canMatch` has benefits of both worlds, it controls whether we can use the route and as a side effect, whether we can download the code. In reality, this means that the chunk won't be loaded if the guard returns `false` but it will also be invoked/triggered every time the user tries to navigate. This is different than `canLoad` which won't load the chunk, but once the guard returned `true` it wouldn't be called again which can cause some undesirable consequences, and `canActivate` which will be called every time but that also meant that chunk will be loaded in all cases. You can read more in the following [blog post](https://netbasal.com/introducing-the-canmatch-router-guard-in-angular-84e398046c9a) or in the [PR description](https://github.com/angular/angular/pull/48180).
```
Erbenos, line 1074: canActivate can still be used, point is that you had to use these in combo before canMatch was implemented to accomplish what you generally want

Erbenos: "This is different than \`canLoad\`" -> "This is different from \`canLoad\`"

Erbenos: "it controls whether we can use the route" -> "it controls when the route can be used" (or when)

update rest of the sentence accordingly

Angular/Angular guidelines and best practices/Formatting, naming and best practices.md
```diff
@@ -1048,6 +1048,31 @@ Some notes:
...
+With Angular v15 it is possible to leverage default exports to shorten the syntax when lazy loading module or a standalone component. Note that the CLI still generates modules and components without the default export so you will need to add that by yourself.
+
+```typescript
+@Component({...})
+export default class LazyCmp { ... }
+
+// Without default export
+{
+  path: 'lazy',
+  loadComponent: () => import('./lazy-file').then(m => m.LazyCmp),
```
Erbenos, line 1062: lazy-file is weird naming, just use index/detail whatever, just something that reflects its a component that represents a page. There is nothing like a "lazy" component

Erbenos (changes requested):

Erbenos:
Tenses in some sentences seem to be all over the place, but we won't sweat it. Alternatively you could run it through some corrector, will leave it up to you

Erbenos (approved):

PR #191, opened 2023-03-15
Initial SSO chapter

DarkoKukovec:
The goal of this chapter is to explain how to work with auth (right now the focus is on sso).
Once we have libs for this, we could also add more specifics for them.

New chapter PR:
https://github.com/infinum/frontend-handbook-private/pull/76

### DIFF
--- a/Auth.md
+++ b/Auth.md
@@ -4,23 +4,23 @@ There are multiple ways of doing auth, but in this chapter, we'll focus on the S
 
 ## Single sign-on (SSO)
 
-Single Sign-On (SSO) is a method of authentication that allows users to access multiple systems and applications with a single set of login credentials. SSO process typically starts when a user attempts to access a protected resource. Instead of prompting the user to enter their login credentials, the application redirects the user to an SSO service, which authenticates the user by checking their login credentials against an identity provider, such as a directory service or a database. Once the user is authenticated, the SSO service will generate a token, which is used to represent the user's identity.
+Single Sign-On (SSO) is a method of authentication that allows users to access multiple systems and applications with a single set of login credentials. The SSO process typically starts when a user attempts to access a protected resource. Instead of prompting the user to enter their login credentials, the application redirects the user to an SSO service, which authenticates the user by checking their login credentials against an identity provider, such as a directory service or a database. Once the user is authenticated, the SSO service will generate a token, which is used to represent the user's identity.
 
-The token is usually a JSON Web Token (JWT) which contains claims that are encoded with the user's identity, as well as other information such as the expiration date of the token. The token is then sent back to the application, which uses it to grant the user access to the protected resource.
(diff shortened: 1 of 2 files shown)


Auth.md
```diff
@@ -0,0 +1,51 @@
...
+
+#### With server-side rendering
+
+When we're working on an app that might do API calls fro the server (e.g. fetching data for the server-side render), we need to have the token available on the server, which means that the cookie is the only option. In this case, we can use the approach the specification calls ["backend for frontend"](https://www.ietf.org/archive/id/draft-ietf-oauth-browser-based-apps-10.html#section-6.2). In general, that means that our backend is the SSO client and it should create a separate session for our client code:
+
+1. Log in with a SSO provider
+2. The backend logic handles the redirect URL, gets the tokens based on the activation code
+3. The backend creates a session for the user and saves the tokens in the session
+4. The backend redirects the user to the client app
+5. Each time an authorized API call needs to be made, it needs to go trough our backend proxy, which will get the tokens, refresh them if necessary and then forward the API call with all th necessary auth headers. Those API calls should also include the CSRF tokens to prevent the CSRF attacks.
```
fvoska, line 47:
```suggestion
5. Each time an authorized API call needs to be made, it needs to go trough our backend proxy, which will get the tokens, refresh them if necessary and then forward the API call with all the necessary auth headers. Those API calls should also include the CSRF tokens to prevent the CSRF attacks.
```

fvoska (approved):
Generally this is a good approach, but we should battle-test it a bit with commonly used FE libraries like [MSAL](https://learn.microsoft.com/en-us/azure/active-directory/develop/msal-overview) which has both [React](https://www.npmjs.com/package/@azure/msal-react) and [Angular](https://www.npmjs.com/package/@azure/msal-angular) flavours.

If we are using AD, we will most probably use MSAL and we would have to see how we can use it with BFF and redirect-based flows.

MSAL also has a popup-based flow, but I guess that is out of the question because BFF flow could only work in a redirect-based flow.

fvoska:
Is there some FE lib that we are ok with using for standard OIDC flows?

I can speak for Angular that, if using AD, MSAL makes most sense. For other providers, we can use [angular-auth-oidc-client](https://www.npmjs.com/package/angular-auth-oidc-client). We used `angular-auth-oidc-client` on CEERIS and there it was actually done as you describe here. We didn't have BFF explicitly, but our Rails backend handled the redirect from SSO provider and then set cookies for frontend. CEERIS is SPA, not SSR, but I think that current setup on CEERIS would work with SSR as well.

We could probably do the same with MSAL, but I am not 100% sure.

DarkoKukovec:
angular-auth-oidc-client is another example of a lib [using localStorage](https://github.com/damienbod/angular-auth-oidc-client/blob/394a85f63d35e1cac1723f6588b29dfe56400d80/projects/angular-auth-oidc-client/src/lib/storage/default-localstorage.service.ts#L2) to store sensitive credentials.

The CEERIS approach is good, but it might not be possible or practical to do this in other cases (e.g. with microservices).

Basically, there is only one client side way to do this securely, and that is with service workers. I didn't find a good lib for this, but I'm working on something for that).

PR #194, opened 2023-05-17
644 update AuthRedirect handbook

isBatak:
Fix https://app.productive.io/1-infinum/projects/1186/tasks/task/4665601?board=224094&filter=NDczNDg4&tour=time-week-layout


React/Recipes/Session Handling.md
```diff
@@ -279,15 +279,19 @@ const AuthRedirect: FC<IAuthRedirectProps> = ({ to, ifFound, condition }) => {
...
+  }, 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  [data, isValidating, error, to, ifFound]);
 
   // this component renders nothing since it is only used to redirect if needed.
   return null;
 };
 ```
 
+> **Note** we added `eslint-disable-next-line react-hooks/exhaustive-deps` to the `useEffect` hook. This is because we don't want to rerender the component if `condition` and `push` are changed. We are aware this is a dangerous thing to do, but in this case, it is a necessary evil. This issue will be resolved when React add finish `useEffectEvent` hook. More about this can be found [here](https://react.dev/learn/separating-events-from-effects#reading-latest-props-and-state-with-effect-events)
```
kristian240, line 293:
```suggestion
> **Note** we added `eslint-disable-next-line react-hooks/exhaustive-deps` to the `useEffect` hook. This is because we don't want to rerender the component if `condition` and `push` are changed. We are aware this is a dangerous thing to do, but in this case, it is a necessary evil. This issue will be resolved when React finishes `useEffectEvent` hook. More about this can be found [here](https://react.dev/learn/separating-events-from-effects#reading-latest-props-and-state-with-effect-events)
```

kristian240 (approved):

DarkoKukovec (approved):

PR #196, opened 2023-07-11
Project structure example addition

veksi:
Going through the react example repo, onboarding, workshops etc. I've asked myself multiple times and had people ask me the same:
"How to structure the forms (how to incorporate them) inside the project structure".

Below, I've added a simple example for the `login` / `register` (`auth`) case which should reflect both:

1. How to handle forms inside the current project structure?
2. Where to put `login` / `register` domain inside the structure (to me makes sense to set both inside the `auth` domain). `ForgotPassword` form and flow would be inside the same structure.

I am open for discussion on how to handle this 2 things (form & auth), or should this even be in the example.

### DIFF
--- a/React/Project structure.md
+++ b/React/Project structure.md
@@ -26,6 +26,11 @@ src
 │   │   │   │   └── HomeHeaderSection.tsx
 │   │   │   └── HomeTodoListSection
 │   │   │       └── HomeTodoListSection.tsx
+│   │   ├── auth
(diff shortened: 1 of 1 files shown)


kristian240 (approved):

isBatak:
I would rather write a specific recipe instead of puluting this chapter with additional context.
For example Session handling recipe https://infinum.com/handbook/frontend/react/recipes/session-handling

PR #204, opened 2024-05-21
Update README and add Remark

kamdubiel:


DarkoKukovec (commented):
This is all useful information, I'm just not sure if it should be in the home page of the handbook, since it will be shown to all visitors.
On one hand, it might encourage someone to contribute, but for most people it will be unnecessary or even confusing.

kamdubiel:
> This is all useful information, I'm just not sure if it should be in the home page of the handbook, since it will be shown to all visitors. On one hand, it might encourage someone to contribute, but for most people it will be unnecessary or even confusing.

Sorry I missed your comment - I didn't think about that and I agree, I'll just remove this info because I cannot think of any other place to put this, but I'll keep this PR for the Remark configuration

DarkoKukovec (approved):

PR #205, opened 2024-06-03
Styleguide revamps

MarkoIvanetic:
Check out commit messages for reasoning.

### DIFF
--- a/Styleguide.md
+++ b/Styleguide.md
@@ -376,43 +376,6 @@ let i;
 let length;
 ```
 
-## [Hoisting](#hoisting)
-
-`var` declarations get hoisted to the top of their scope; their assignment does not. `const` and `let` declarations have [Temporal Dead Zones (TDZ)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#Temporal_dead_zone_and_errors_with_let). It's important to know why [typeof is no longer safe](http://es-discourse.com/t/why-typeof-is-no-longer-safe/15).
-
-```js
-// we know this wouldn't work (assuming there
-// is no notDefined global variable)
-function example() {
-  console.log(notDefined); // => throws a ReferenceError
-}
-
-// creating a variable declaration after you
-// reference the variable will work due to
-// variable hoisting. Note: the assignment
-// value of `true` is not hoisted.
-function example() {
-  console.log(declaredButNotAssigned); // => undefined
-  var declaredButNotAssigned = true;
-}
-
-// the interpreter is hoisting the variable
-// declaration to the top of the scope,
-// which means our example could be rewritten as:
-function example() {
-  let declaredButNotAssigned;
-  console.log(declaredButNotAssigned); // => undefined
-  declaredButNotAssigned = true;
-}
-
-// using const and let
-function example() {
-  console.log(declaredButNotAssigned); // => throws a ReferenceError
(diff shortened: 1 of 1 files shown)


Styleguide.md
```diff
@@ -39,15 +40,15 @@ function getKey(k) {
 // bad
 const obj = {
   id: 5,
-  name: 'San Francisco',
+  name: "San Francisco",
```
DarkoKukovec, line 43: How come there are double quotes? Shouldn't single quotes be preferred based on our prettier config?

Styleguide.md
```diff
@@ -190,17 +193,17 @@ When programmatically building up strings, use template strings instead of conca
...
 // bad
 function sayHi(name) {
-  return ['How are you, ', name, '?'].join();
+  return ["How are you, ", name, "?"].join();
 }
 
 // bad
 function sayHi(name) {
-  return `How are you, ${ name }?`;
+  return `How are you, ${name}?`;
```
DarkoKukovec, line 206: How is this different from the good example?

Styleguide.md
```diff
@@ -350,170 +351,154 @@ export default es6;
...
 
 // good
-function checkName(hasName) {
-  if (hasName === 'test') {
-    return false;
-  }
+const heroes = ["Batman", "Superman"];
+```
 
-  const name = getName();
```
DarkoKukovec, line 401: Did you remove this in the end? Regarding the ordering, I guess I agree, but it would depend on the `getName` call complexity if it should be done before the if or just as a let before, and the calculation if needed.

Styleguide.md
fvoska: This repo doesn't have prettier config, so whatever are author's default formatter settings in editor/IDE, that will end up happening. By default, it might not even be using prettier as formatter, but VSCode's default formatter.

Easy fix should be to add prettierrc to this repo. @MarkoIvanetic can you please do that?

kamdubiel: The problem with adding prettier configuration is that it will automatically fix some of the "bad" code.

I think that Marko had prettier installed globally, and this document was formatted with Prettier's default config since there's no prettier specific config in the repo. IMO we should disable prettier for this repository with:
```
// .prettierignore
**
```
instead of adding .prettierrc file

kamdubiel: Second thought, I added Remark configuration in the other PR, let me know what you think:
https://github.com/infinum/frontend-handbook/pull/204/commits/56c84eda779a91e44dc001ae206f98a000154a17

fvoska: Seems like a good option, I forgot about the fact that prettier would format bad examples as well.

It does however highlight how our handbook has some recommendations about formatting that are probably unnecessary, as prettier will be used on most projects. We could trim the fat a bit by removing formating-only recommendations from handbook

Styleguide.md
MarkoIvanetic: Do you want this example returned and refactored? @DarkoKukovec

Styleguide.md
MarkoIvanetic: disabled my global prettier and had no issues after. remark is a great option 👍

Styleguide.md
MarkoIvanetic: removed

DarkoKukovec (approved):

PR #208, opened 2024-06-25
Update Managing versions part

kamdubiel:


Node/Package managers guidelines.md
```diff
@@ -0,0 +1,200 @@
...
+
+#### Installing Dependencies
+
+To install all dependencies listed in your package.json, run:
+
+```bash
+pnpm install
+```
+
+#### Adding a New Package
```
safo6m, line 131: Maybe describe some frequently used flags, like `prod` and `frozen-lockfile`.

kamdubiel: Added more commands

Node/Package managers guidelines.md
```diff
@@ -0,0 +1,282 @@
...
+To install packages from the local cache only, without trying to reach the internet:
+
+```bash
+pnpm install --offline
+```
+
+You can also combine multiple flags together, which is useful in CI/CD, for example:
+
+```bash
+pnpm install --prod --offline --frozen-lockfile
```
safo6m, line 186: `--offline` shouldn't be used with CI/CD. 
I don't think you suggest that, but as a CI/CD is mentioned in the sentence above, it could be interpreted as such.

safo6m (changes requested):

safo6m (approved):

thabalija (approved):
Very nice and detailed explanation 👏

